PR #2, opened 2025-01-02
Update module name, update go.sum

simonlechner:
During the change of the module name, the go.sum file has not been updated correctly.
The linting issues have been resolved.


simonlechner:
The update of module names is already a very extensive change, removing the dot imports should not happen in the same PR.

PR #5, opened 2025-01-10
CT: No code is treated as STOP

simonlechner:
`returnData` is write only, therefore it should not be set in the CT state generation. RETURNDATASIZE and RETURNDATACOPY opcodes are reading from `lastCallReturnData`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -85,6 +85,6 @@ require (
 	rsc.io/tmplfunc v0.0.3 // indirect
 )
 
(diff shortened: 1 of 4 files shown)


HerbertJordan (approved):
I assume this is identical to [#958](https://github.com/Fantom-foundation/Tosca/pull/958)

simonlechner:
> I assume this is identical to [#958](https://github.com/Fantom-foundation/Tosca/pull/958)

Yes, it is identical except for the updated 0xsoniclabs/go-ethereum reference in the go.mod file.

HerbertJordan (approved):

PR #7, opened 2025-01-12
evmrs: let gas_refund overflow to pass regressions

LorenzSchueler:
Unlike for gas_left, it seems to be expected that gas_refund DOES overflow.
This PR removed the overflow check which makes the regression tests pass.

### DIFF
--- a/rust/src/interpreter.rs
+++ b/rust/src/interpreter.rs
@@ -1419,7 +1419,7 @@ impl<const STEPPABLE: bool> Interpreter<'_, STEPPABLE> {
 
         let destructed = self.context.selfdestruct(self.message.recipient(), &addr);
         if self.revision <= Revision::EVMC_BERLIN && destructed {
-            self.gas_refund.add(24_000)?;
+            self.gas_refund.add(24_000);
         }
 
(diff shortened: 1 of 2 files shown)


LorenzSchueler:
CI should pass once #6 is merged.

rust/src/interpreter.rs
```diff
@@ -1419,7 +1419,7 @@ impl<const STEPPABLE: bool> Interpreter<'_, STEPPABLE> {
 
         let destructed = self.context.selfdestruct(self.message.recipient(), &addr);
         if self.revision <= Revision::EVMC_BERLIN && destructed {
-            self.gas_refund.add(24_000)?;
+            self.gas_refund.add(24_000);
```
LuisPH3, line 1422: I am afraid this is ignoring a result type, and linters will not like it.
can you make explicit that result is ignored? the reason why explanation can be found via blame...

LuisPH3 (changes requested):

LuisPH3 (approved):
my bad, missed rewrite of add function

simonlechner (approved):
I like the `?` operator.

PR #8, opened 2025-01-12
evmrs: adjust code cache size depending on enabled features

LorenzSchueler:
The code analysis size (in the number of entries) was taken from evmzero. However, some features blow up the size of the internal representation of each entry. This leads to too much memory consumption. This PR therefore decreases the number of entries in those cases to keep the cache size in bytes roughly the same.


LorenzSchueler:
Ci should pass once #6 is merged.

LuisPH3 (approved):

PR #9, opened 2025-01-13
evmrs: use push guard to avoid stack overflow checks

LorenzSchueler:
This PR adds the method `pop_with_guard` to Stack, which works similar to the normal `pop` method, but additionally returns a PushGuard which holds a mutable reference to the location in the stack where the result can be pushed to. Because we popped values first, pushing a single value will never overflow the stack. 
Unfortunately, when first poping values and then pushing a single value, the compiler does not figure this out and adds the check for the (evm)stack overflow. With this optimization, this check is now gone.
While this optimization does decrease performance on old hardware (probably because of code layout reasons), it does improve performance on new hardware, which is what matters at the end of the day.

```
goos: linux
goarch: amd64
pkg: github.com/Fantom-foundation/Tosca/go/integration_test/interpreter
cpu: AMD EPYC 7713 64-Core Processor                
                   │ 2025-01-08T09:35#4da02ec#20-main/evmrs#performance │ 2025-01-09T14:11#2574774#20/evmrs#performance │
                   │                       sec/op                       │         sec/op          vs base               │
StaticOverhead/1/                                           900.9n ± 0%              897.4n ± 0%       ~ (p=0.106 n=20)
Inc/1/                                                      1.912µ ± 1%              1.904µ ± 0%  -0.42% (p=0.031 n=20)
Inc/10/                                                     1.918µ ± 0%              1.862µ ± 0%  -2.92% (p=0.000 n=20)
Fib/1/                                                      1.688µ ± 1%              1.646µ ± 0%  -2.49% (p=0.000 n=20)
Fib/5/                                                      8.405µ ± 0%              8.308µ ± 4%       ~ (p=0.286 n=20)
Fib/10/                                                     89.09µ ± 0%              89.13µ ± 0%       ~ (p=0.507 n=20)
Fib/15/                                                     931.9µ ± 1%              924.7µ ± 1%  -0.77% (p=0.020 n=20)
Fib/20/                                                     10.36m ± 1%              10.20m ± 1%  -1.47% (p=0.001 n=20)
Sha3/1/                                                     1.088µ ± 0%              1.063µ ± 0%  -2.25% (p=0.000 n=20)
Sha3/10/                                                    1.837µ ± 0%              1.800µ ± 0%  -2.01% (p=0.000 n=20)
Sha3/100/                                                   9.357µ ± 0%              9.217µ ± 0%  -1.51% (p=0.000 n=20)
Sha3/1000/                                                  87.29µ ± 0%              85.71µ ± 0%  -1.81% (p=0.000 n=20)
Arith/1/                                                    2.081µ ± 0%              2.189µ ± 1%  +5.19% (p=0.000 n=20)
Arith/10/                                                   4.841µ ± 1%              4.841µ ± 4%       ~ (p=0.213 n=20)
Arith/100/                                                  32.08µ ± 0%              31.68µ ± 0%  -1.23% (p=0.000 n=20)
Arith/280/                                                 101.27µ ± 7%              91.41µ ± 1%  -9.74% (p=0.000 n=20)
Memory/1/                                                   2.886µ ± 1%              2.799µ ± 0%  -3.02% (p=0.000 n=20)
Memory/10/                                                  6.193µ ± 0%              6.096µ ± 1%  -1.56% (p=0.000 n=20)
Memory/100/                                                 39.37µ ± 0%              36.71µ ± 0%  -6.76% (p=0.000 n=20)
Memory/1000/                                                373.5µ ± 0%              366.3µ ± 0%  -1.94% (p=0.000 n=20)
Memory/10000/                                               3.617m ± 0%              3.347m ± 6%  -7.47% (p=0.000 n=20)
Analysis/jumpdest/                                          927.1n ± 0%              913.4n ± 0%  -1.47% (p=0.000 n=20)
Analysis/stop/                                              939.8n ± 0%              917.4n ± 0%  -2.38% (p=0.000 n=20)
Analysis/push1/                                             937.2n ± 1%              917.1n ± 0%  -2.15% (p=0.000 n=20)
Analysis/push32/                                            945.2n ± 1%              912.9n ± 0%  -3.42% (p=0.000 n=20)
geomean                                                     12.24µ                   11.98µ       -2.16%
```

### DIFF
--- a/rust/src/types/stack.rs
+++ b/rust/src/types/stack.rs
@@ -10,9 +10,16 @@ impl<const N: usize> NonZero<N> {
     const VALID: () = assert!(N > 0);
 }
 
-/// Wrapper around [`&mut u256`] that ensures that the only possible operation is to write once to
-/// this memory location.
+/// This type is created by calling [`Stack::pop_with_location`] and is intended to replace pushing
+/// to the stack directly. It and avoids the stack overflow check when pushing because it is no
+/// longer needed. [`PushLocation`] has to be consumed by pushing to it.
+/// If this does not happen, the program is still memory safe, however there will be one item so
+/// much on the stack.
+///
+/// Internally it is a wrapper around [`&mut u256`] that ensures that the only possible operation is
+/// to write once to this memory location.
 #[derive(Debug)]
+#[must_use = "PushLocation has to be pushed to."]
 pub struct PushLocation<'p>(&'p mut u256);
 
 impl PushLocation<'_> {
--- a/rust/src/types/stack.rs
+++ b/rust/src/types/stack.rs
@@ -12,9 +12,9 @@ impl<const N: usize> NonZero<N> {
 
 /// Wrapper around [`&mut u256`] that ensures that the only possible operation is to write once to
 /// this memory location.
-pub struct PushGuard<'p>(&'p mut u256);
+pub struct PushLocation<'p>(&'p mut u256);
 
-impl PushGuard<'_> {
+impl PushLocation<'_> {
     pub fn push(self, value: impl Into<u256>) {
         *self.0 = value.into();
     }
@@ -124,7 +124,9 @@ impl Stack {
         Ok(array)
     }
 
-    pub fn pop_with_guard<const N: usize>(&mut self) -> Result<(PushGuard, [u256; N]), FailStatus> {
+    pub fn pop_with_location<const N: usize>(
+        &mut self,
+    ) -> Result<(PushLocation, [u256; N]), FailStatus> {
         self.check_underflow(N)?;
 
         self.0.truncate(self.len() - (N - 1));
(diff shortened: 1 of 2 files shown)


rust/src/types/stack.rs
```diff
@@ -10,6 +10,16 @@ impl<const N: usize> NonZero<N> {
     const VALID: () = assert!(N > 0);
 }
 
+/// Wrapper around [`&mut u256`] that ensures that the only possible operation is to write once to
+/// this memory location.
```
LuisPH3, line 14: I am not convinced about the name: guard makes me think that this is a raii object, but requires a push to happen.
Regarding the documentation string. instead of describing how it is implemented (a wrapper over a slice) it would be better to describe what is the intended use and how to use it.

What happens if developper forgets to call push?

LuisPH3 (approved):
:+1:

PR #10, opened 2025-01-13
no bounds checks in get_push_data

LorenzSchueler:
This PR adds unsafe compiler hints to remove bound checks.
The idea is that when feature fn-ptr-conversion-expanded-dispatch is enabled, the analysis item holds both the function pointer for the corresponding opcode impl and a u256 used as push data.
The function pointer is retrieved first in order to call the function.
In case the function is Interpreter::push, it calls get_push_data to get the data to push onto the stack.
Because we already retrieved the function pointer, and the pc was not modified in between, it is not necessary to check the bounds again when accessing the same analysis item again to get the push data.


LuisPH3:
It is a good compromise.

PR #12, opened 2025-01-14
evmrs: add feature unsafe-hints

LorenzSchueler:
This PR adds more unsafe compiler hints to eliminate bound checks and puts those behind the new feature `unsafe-hints`.

- assert that the slice in m_store is 32 bytes long
- assert that the memory access after memory expansion is always in bounds
- put the existing hint of #10 also behind the new feature

Additionally, the part of the memory expansion function which is responsible for the actual expansion (not the checks if an expansion is needed) is extracted into a separate function so that the checks without the expansions get inlined.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -27,6 +27,7 @@ needs-cache = ["dep:lru"]
 performance = [
     "mimalloc",
     "unsafe-stack",
+    "unsafe-hints",
     "custom-evmc",
     "hash-cache",
     "code-analysis-cache",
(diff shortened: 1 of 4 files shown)


LuisPH3:
A link to unsafe-hints would be nice to have, because this is a low level feature of rust, not everyone may know about it.

LuisPH3 (commented):
Question: Are this hints based on the previous code generation? or speculated?

LorenzSchueler:
By adding those hints, the programmer guarantees that the mentioned conditions always hold. The compiler is then allowed to optimize the code by taking this information into account without having to check it. This is why it is unsafe, because if the conditions do not hold, UB might occur.
As the docs mention, this is equivalen to `[__builtin_assume](https://clang.llvm.org/docs/LanguageExtensions.html#builtin-assume).` in C.

LuisPH3 (approved):

LorenzSchueler:
The specific locations for where to add those hints were found by profiling and then looking at the assembly of functions which had a high higher than expected execution time.

PR #14, opened 2025-01-15
evmrs: parity between rust benchmarks and go vm_test benchmarks

LorenzSchueler:
This PR creates parity between the Rust and Go benchmarks.
On the Rust side the missing benchmark sizes are added.
On the Go side the StaticOverhead benchmark is changed to be identical to the Rust version.


LuisPH3 (approved):
:+1:

PR #15, opened 2025-01-17
Geth processor reference implementation

simonlechner:
For the release of the Sonic network the processor implementation has been updated to the one from ethereum with only minor changes. To be able to test Floria against this new implementation, it is added as a processor option to tosca.

Tests to check this integration:
- [x] Tosca processor integration tests
- [x] Fantom mainnet history

### DIFF
--- a/.testcoverage.yml
+++ b/.testcoverage.yml
@@ -92,6 +92,8 @@ override:
   - threshold: 0
     path: go/processor/floria
   - threshold: 0
+    path: go/processor/geth
+  - threshold: 0
     path: go/processor/opera
   - threshold: 0
     path: go/tosca
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -29,7 +29,7 @@ import (
 )
 
 func init() {
-	tosca.RegisterProcessorFactory("geth-ftm", fantomProcessor)
+	tosca.RegisterProcessorFactory("geth-sonic", fantomProcessor)
 }
 
 func fantomProcessor(interpreter tosca.Interpreter) tosca.Processor {
@@ -98,7 +98,7 @@ func newBlockContext(blockParameters tosca.BlockParameters, context tosca.Transa
 		return common.Hash(context.GetBlockHash(int64(num)))
 	}
 
-	ftmDifficulty := big.NewInt(1)
+	sonicDifficulty := big.NewInt(1)
 
 	return vm.BlockContext{
 		CanTransfer: canTransfer,
@@ -108,7 +108,7 @@ func newBlockContext(blockParameters tosca.BlockParameters, context tosca.Transa
 		GasLimit:    uint64(blockParameters.GasLimit),
 		BlockNumber: new(big.Int).SetInt64(blockParameters.BlockNumber),
 		Time:        uint64(blockParameters.Timestamp),
-		Difficulty:  ftmDifficulty,
+		Difficulty:  sonicDifficulty,
 		BaseFee:     blockParameters.BaseFee.ToBig(),
 		BlobBaseFee: blockParameters.BlobBaseFee.ToBig(),
 		Random:      (*common.Hash)(&blockParameters.PrevRandao),
@@ -229,7 +229,9 @@ func (s *stateDB) AddBalance(address common.Address, value *uint256.Int, tracing
 	s.context.SetBalance(toscaAddress, tosca.Add(balance, tosca.ValueFromUint256(value)))
 
 	// In the case of a seldestruct the balance is transferred to the beneficiary,
-	// we save this address for the context-selfdestruct call
+	// we save this address for the context-selfdestruct call.
+	// this only works if the balance transfer is performed before the selfdestruct call,
+	// as it is the performed in geth and the geth adapter.
 	s.beneficiary = address
 }
 
@@ -322,10 +324,12 @@ func (s *stateDB) Empty(address common.Address) bool {
 }
 
 func (s *stateDB) AddressInAccessList(address common.Address) bool {
+	// using the non-deprecated function has a side effect of adding the address to the access list
 	return bool(s.context.AccessAccount(tosca.Address(address)))
 }
 
 func (s *stateDB) SlotInAccessList(address common.Address, slot common.Hash) (addressOk bool, slotOk bool) {
+	// using the non-deprecated functions has a side effect of adding the address and slot to the access list
 	addressOk = bool(s.context.AccessAccount(tosca.Address(address)))
 	slotOk = bool(s.context.AccessStorage(tosca.Address(address), tosca.Key(slot)))
 	return addressOk, slotOk
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -57,7 +57,7 @@ func (p *Processor) Run(
 	}
 	stateDB := &stateDB{context: context}
 	chainConfig := blockParametersToChainConfig(blockParameters)
-	config := evmConfig(p.interpreter, p.ethereumCompatible)
+	config := newEVMConfig(p.interpreter, p.ethereumCompatible)
 	evm := vm.NewEVM(blockContext, txContext, stateDB, chainConfig, config)
 
 	msg := transactionToMessage(transaction, blockParameters.BaseFee)
@@ -98,6 +98,8 @@ func newBlockContext(blockParameters tosca.BlockParameters, context tosca.Transa
 		return common.Hash(context.GetBlockHash(int64(num)))
 	}
 
+	ftmDifficulty := big.NewInt(1)
+
 	return vm.BlockContext{
 		CanTransfer: canTransfer,
 		Transfer:    transfer,
@@ -106,7 +108,7 @@ func newBlockContext(blockParameters tosca.BlockParameters, context tosca.Transa
 		GasLimit:    uint64(blockParameters.GasLimit),
 		BlockNumber: new(big.Int).SetInt64(blockParameters.BlockNumber),
 		Time:        uint64(blockParameters.Timestamp),
-		Difficulty:  big.NewInt(1),
+		Difficulty:  ftmDifficulty,
 		BaseFee:     blockParameters.BaseFee.ToBig(),
 		BlobBaseFee: blockParameters.BlobBaseFee.ToBig(),
 		Random:      (*common.Hash)(&blockParameters.PrevRandao),
@@ -125,30 +127,28 @@ func blockParametersToChainConfig(blockParams tosca.BlockParameters) *params.Cha
 	chainConfig.ShanghaiTime = &zeroTime
 	chainConfig.CancunTime = &zeroTime
 
+	greaterBlockTime := uint64(blockParams.Timestamp + 1)
+	greaterBlockNumber := big.NewInt(blockParams.BlockNumber + 1)
+
 	if blockParams.Revision < tosca.R13_Cancun {
-		time := uint64(blockParams.Timestamp + 1)
-		chainConfig.CancunTime = &time
+		chainConfig.CancunTime = &greaterBlockTime
 	}
 	if blockParams.Revision < tosca.R12_Shanghai {
-		time := uint64(blockParams.Timestamp + 1)
-		chainConfig.ShanghaiTime = &time
+		chainConfig.ShanghaiTime = &greaterBlockTime
 	}
 	if blockParams.Revision < tosca.R11_Paris {
-		chainConfig.MergeNetsplitBlock = big.NewInt(blockParams.BlockNumber + 1)
+		chainConfig.MergeNetsplitBlock = greaterBlockNumber
 	}
 	if blockParams.Revision < tosca.R10_London {
-		chainConfig.LondonBlock = big.NewInt(blockParams.BlockNumber + 1)
+		chainConfig.LondonBlock = greaterBlockNumber
 	}
 	if blockParams.Revision < tosca.R09_Berlin {
-		chainConfig.BerlinBlock = big.NewInt(blockParams.BlockNumber + 1)
-	}
-	if blockParams.Revision < tosca.R07_Istanbul {
-		chainConfig.IstanbulBlock = big.NewInt(blockParams.BlockNumber + 1)
+		chainConfig.BerlinBlock = greaterBlockNumber
 	}
 	return &chainConfig
 }
 
-func evmConfig(interpreter tosca.Interpreter, ethereumCompatible bool) vm.Config {
+func newEVMConfig(interpreter tosca.Interpreter, ethereumCompatible bool) vm.Config {
 	config := vm.Config{
 		StatePrecompiles: map[common.Address]vm.PrecompiledStateContract{
 			stateContractAddress: PreCompiledContract{},
@@ -185,10 +185,12 @@ func transactionToMessage(transaction tosca.Transaction, baseFee tosca.Value) *c
 		GasLimit: uint64(transaction.GasLimit),
 		GasPrice: transaction.GasPrice.ToBig(),
 		// gas price computation and enforcement of cap is currently performed outside of processor
-		GasFeeCap:         big.NewInt(0).Add(baseFee.ToBig(), big.NewInt(1)),
-		GasTipCap:         big.NewInt(0),
-		Data:              transaction.Input,
-		AccessList:        accessList,
+		// TODO: extend the tosca.Transaction to include GasFeeCap and GasTipCap
+		GasFeeCap:  big.NewInt(0).Add(baseFee.ToBig(), big.NewInt(1)),
+		GasTipCap:  big.NewInt(0),
+		Data:       transaction.Input,
+		AccessList: accessList,
+		// TODO: add support for blobs in the tosca.Transaction
 		BlobGasFeeCap:     big.NewInt(0),
 		BlobHashes:        nil,
 		SkipAccountChecks: false,
@@ -196,6 +198,7 @@ func transactionToMessage(transaction tosca.Transaction, baseFee tosca.Value) *c
 }
 
 // stateDB is a wrapper around the tosca.TransactionContext to implement the vm.StateDB interface.
+// TODO: merge with tosca/go/interpreter/geth/geth.go stateDbAdapter
 type stateDB struct {
 	context         tosca.TransactionContext
 	refund          uint64
@@ -225,7 +228,8 @@ func (s *stateDB) AddBalance(address common.Address, value *uint256.Int, tracing
 	balance := s.context.GetBalance(toscaAddress)
 	s.context.SetBalance(toscaAddress, tosca.Add(balance, tosca.ValueFromUint256(value)))
 
-	// we save this address to be used as the beneficiary in a selfdestruct case.
+	// In the case of a seldestruct the balance is transferred to the beneficiary,
+	// we save this address for the context-selfdestruct call
 	s.beneficiary = address
 }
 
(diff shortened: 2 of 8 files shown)


go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	if blockParams.Revision < tosca.R13_Cancun {
+		time := uint64(blockParams.Timestamp + 1)
+		chainConfig.CancunTime = &time
+	}
+	if blockParams.Revision < tosca.R12_Shanghai {
+		time := uint64(blockParams.Timestamp + 1)
+		chainConfig.ShanghaiTime = &time
+	}
+	if blockParams.Revision < tosca.R11_Paris {
+		chainConfig.MergeNetsplitBlock = big.NewInt(blockParams.BlockNumber + 1)
```
facuMH, line 137: same here, `big.NewInt(blockParams.BlockNumber + 1)` could be reused. but neither this nor `time` are critical comments.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+		Data:              transaction.Input,
+		AccessList:        accessList,
+		BlobGasFeeCap:     big.NewInt(0),
+		BlobHashes:        nil,
+		SkipAccountChecks: false,
+	}
+}
+
+// stateDB is a wrapper around the tosca.TransactionContext to implement the vm.StateDB interface.
+type stateDB struct {
```
facuMH, line 199: what  do you think about `fmtStateDB` ?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	refund          uint64
+	createdContract common.Address
+	refundBackups   map[tosca.Snapshot]uint64
+	beneficiary     common.Address
+}
+
+// vm.StateDB interface implementation
+
+func (s *stateDB) CreateAccount(common.Address) {
+	// not implemented
```
facuMH, line 210: do we want this to panic ? I believe this implementation is just for testing, no ?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	s.refund -= refund
+}
+
+func (s *stateDB) GetRefund() uint64 {
+	return s.refund
+}
+
+func (s *stateDB) GetCommittedState(address common.Address, key common.Hash) common.Hash {
+	//lint:ignore SA1019 deprecated functions to be migrated
+	return common.Hash(s.context.GetCommittedStorage(tosca.Address(address), tosca.Key(key)))
```
facuMH, line 274: why not use the new function instead ?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	toscaLog := tosca.Log{
+		Address: tosca.Address(log.Address),
+		Topics:  topics,
+		Data:    log.Data,
+	}
+	s.context.EmitLog(tosca.Log(toscaLog))
+}
+
+func (s *stateDB) AddPreimage(common.Hash, []byte) {
+	panic("not implemented")
```
facuMH, line 392: it seems a little inconsist that some functions get a comment and other a panic. is there a reason for these ?

go/processor/geth/state_contract.go
```diff
@@ -0,0 +1,216 @@
...
+		// setBalance
+		if suppliedGas < params.CallValueTransferGas {
+			return nil, 0, vm.ErrOutOfGas
+		}
+		suppliedGas -= params.CallValueTransferGas
+		if len(input) != 64 {
+			return nil, 0, vm.ErrExecutionReverted
+		}
+
+		acc := common.BytesToAddress(input[12:32])
```
facuMH, line 88: what are bytes 4 to 11 ?

go/processor/geth/state_contract.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+
+		balance := stateDB.GetBalance(acc)
+		if balance.Cmp(value) >= 0 {
+			diff := new(uint256.Int).Sub(balance, value)
+			stateDB.SubBalance(acc, diff, tracing.BalanceChangeUnspecified)
+		} else {
+			diff := new(uint256.Int).Sub(value, balance)
+			stateDB.AddBalance(acc, diff, tracing.BalanceChangeUnspecified)
+		}
```
facuMH, line 104:
```suggestion
		if balance.Cmp(value) >= 0 {
			diff := new(uint256.Int).Sub(balance, value)
		} else {
			diff := new(uint256.Int).Sub(value, balance)
		}
		stateDB.AddBalance(acc, diff, tracing.BalanceChangeUnspecified)
```

go/processor/geth/state_contract.go
```diff
@@ -0,0 +1,216 @@
...
+
+		*constID = make([]byte, len(method.ID))
+		copy(*constID, method.ID)
+	}
+}
+
+type PreCompiledContract struct{}
+
+func (PreCompiledContract) Run(stateDB vm.StateDB, _ vm.BlockContext, txCtx vm.TxContext, caller common.Address, input []byte, suppliedGas uint64) ([]byte, uint64, error) {
```
facuMH, line 70: nit: there is repeated code in the different branches of the if that could be abstracted.

facuMH (commented):
It looks fine to me. no semantic errors that I could spot. That said I have some coding style comments and questions.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	"github.com/ethereum/go-ethereum/core/stateless"
+	"github.com/ethereum/go-ethereum/core/tracing"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/trie/utils"
+)
+
+func init() {
+	tosca.RegisterProcessorFactory("geth-ftm", fantomProcessor)
```
HerbertJordan, line 32: Should we rename this processor to "sonic" and the factory to `sonicProcessor`?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	stateDB := &stateDB{context: context}
+	chainConfig := blockParametersToChainConfig(blockParameters)
+	config := evmConfig(p.interpreter, p.ethereumCompatible)
+	evm := vm.NewEVM(blockContext, txContext, stateDB, chainConfig, config)
+
+	msg := transactionToMessage(transaction, blockParameters.BaseFee)
+	gasPool := new(core.GasPool).AddGas(uint64(transaction.GasLimit))
+	result, err := core.ApplyMessage(evm, msg, gasPool)
+	if err != nil {
+		if !p.ethereumCompatible && errors.Is(err, core.ErrInsufficientFunds) {
```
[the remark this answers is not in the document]
simonlechner: In fantom it is not an error if the topmost call does not have sufficient funds, the transaction returns an unsuccessful receipt but the gas is not consumed. If the `ethereumCompatible` flag is set, also the `InsufficientBalanceIsNotAnError`.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+			return tosca.Receipt{}, err
+		}
+		return tosca.Receipt{GasUsed: transaction.GasLimit}, err
```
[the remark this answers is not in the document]
simonlechner: Yes

.testcoverage.yml
```diff
@@ -91,6 +91,8 @@ override:
     path: go/interpreter/lfvm
   - threshold: 0
     path: go/processor/floria
+  - threshold: 0
+    path: go/processor/geth
```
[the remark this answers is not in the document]
simonlechner: The majority of the code is just passing values to the geth implementation, which I think is a little bit hard to test on an unit tests level. If we add the integration tests to the coverage it is greater 88%. 
If you have any  ideas for concrete unit tests, please let me know.

go/processor/geth/state_contract.go
HerbertJordan, line 1: This looks like it is copied from the Sonic repository. Could we have a dependency instead?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	if blockParams.Revision < tosca.R09_Berlin {
+		chainConfig.BerlinBlock = big.NewInt(blockParams.BlockNumber + 1)
+	}
+	if blockParams.Revision < tosca.R07_Istanbul {
+		chainConfig.IstanbulBlock = big.NewInt(blockParams.BlockNumber + 1)
+	}
+	return &chainConfig
+}
+
+func evmConfig(interpreter tosca.Interpreter, ethereumCompatible bool) vm.Config {
```
HerbertJordan, line 151: Function names should start with a verb.

go/processor/geth/processor.go
simonlechner: It is called during the tests and history run, so it can not panic.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+	return &core.Message{
+		From:     common.Address(transaction.Sender),
+		To:       (*common.Address)(transaction.Recipient),
+		Nonce:    transaction.Nonce,
+		Value:    transaction.Value.ToBig(),
+		GasLimit: uint64(transaction.GasLimit),
+		GasPrice: transaction.GasPrice.ToBig(),
+		// gas price computation and enforcement of cap is currently performed outside of processor
+		GasFeeCap:         big.NewInt(0).Add(baseFee.ToBig(), big.NewInt(1)),
+		GasTipCap:         big.NewInt(0),
```
HerbertJordan, line 189: I guess this should not remain 0 - maybe add a TODO as a reminder?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+		Data:              transaction.Input,
+		AccessList:        accessList,
+		BlobGasFeeCap:     big.NewInt(0),
```
HerbertJordan, line 192: Same here?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+		BlobHashes:        nil,
```
HerbertJordan, line 193: And here?

go/processor/geth/state_contract.go
simonlechner: This file is copied from opera (I added a comment), I don't want to do any changing on it to ensure it is conformant with the version running in production. 
4 to 11 is just padding.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+func (s *stateDB) CreateContract(address common.Address) {
+	s.createdContract = address
+}
+
+func (s *stateDB) SubBalance(address common.Address, value *uint256.Int, tracing tracing.BalanceChangeReason) {
+	toscaAddress := tosca.Address(address)
+	balance := s.context.GetBalance(toscaAddress)
+	s.context.SetBalance(toscaAddress, tosca.Sub(balance, tosca.ValueFromUint256(value)))
```
HerbertJordan, line 220: Should there be a check that the available balance is higher than the balance to be subtracted?

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+}
+
+func (s *stateDB) AddBalance(address common.Address, value *uint256.Int, tracing tracing.BalanceChangeReason) {
+	toscaAddress := tosca.Address(address)
+	balance := s.context.GetBalance(toscaAddress)
+	s.context.SetBalance(toscaAddress, tosca.Add(balance, tosca.ValueFromUint256(value)))
+
+	// we save this address to be used as the beneficiary in a selfdestruct case.
+	s.beneficiary = address
```
HerbertJordan, line 229: I assume this only works in cooperation with an interpreter wrapped into a geth adapter? (`geth_adapter.NewGethInterpreterFactory(...)`); If so, it would be nice to explain this mechanism here.

go/processor/geth/processor.go
simonlechner: I want all functions for which is possible to panic to panic as this will save a lot of debugging time once they are required.

go/processor/geth/processor.go
HerbertJordan, line 199: This type is serving the same purpose as the `stateDbAdapter` [here](https://github.com/Fantom-foundation/Tosca/blob/4da02eca5a6c16bfcec36c1c401957cab71b8e57/go/interpreter/geth/geth.go#L220). It would be great to merge those implementations.

There seem to be additional synergies between the `geth` interpreter package and this package that should be utilized. Wdyt?

HerbertJordan (changes requested):
It seems like this PR might be introducing some code duplication. I added some remarks about potential redundancy reduction options. Please have a look.

go/processor/geth/processor.go
simonlechner: I changed the access list related functions to the new functions. For the self destruct and committed state this is not so straight forward so I will leave them with the outdated functions for now.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,397 @@
...
+
+type Processor struct {
+	interpreter        tosca.Interpreter
+	ethereumCompatible bool
+}
+
+func (p *Processor) Run(
+	blockParameters tosca.BlockParameters,
+	transaction tosca.Transaction,
+	context tosca.TransactionContext,
```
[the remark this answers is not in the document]
simonlechner: There is already a `txContext` in this block. I don't see where it is overloaded, where else is context used?

go/processor/geth/processor.go
simonlechner: I have a follow up PR that extends the implementation to be ethereum compatible. Since this will then fully match the geth implementation I would prefer to also call it geth. I think it also makes sense to have a geth interpreter and processor.

go/processor/geth/state_contract.go
simonlechner: Tosca is a dependency in Sonic, therefore Sonic can not be one in Tosca.

go/processor/geth/processor.go
simonlechner: This function is only called after a check for sufficient balance has been performed. I can not add an error due to the interface and I don't want the function to panic and kill the whole process.

go/processor/geth/processor.go
simonlechner: Yes there are a few places where code could be reduced.
I would prefer to move this to a separate follow up PR to keep PRs small and keep additions in this PR to the geth processor. I added a TODO in the code.

simonlechner: I don't see anything wrong with stateDB and we should not add ftm to something that is not fantom specific.

go/processor/geth/processor.go
simonlechner: Yes, these fields should be updated with the update to the `tosca.Transaction`, I added TODOs.

go/processor/geth/processor.go
simonlechner: It should not matter whether the interpreter is wrapped in the `geth_adapter` or not, no matter how it is integrated there has to be a balance update before selfdestruct is called in the stateDB. For example the geth interpreter is not wrapped by the adapter but performs the balance update in the instruction implementation of selfdestruct. 
I updated the comment in the `AddBalance` function.

go/processor/geth/processor.go
```diff
@@ -318,13 +322,13 @@ func (s *stateDB) Empty(address common.Address) bool {
 }
 
 func (s *stateDB) AddressInAccessList(address common.Address) bool {
-	//lint:ignore SA1019 deprecated functions to be migrated
-	return s.context.IsAddressInAccessList(tosca.Address(address))
+	return bool(s.context.AccessAccount(tosca.Address(address)))
```
HerbertJordan, line 325: Note: this has the side-effect that the address is also added to the access list, not just checking whether it is there.

So this is only true if a call of this function is immediately followed by a `AddAddressToAccessList` -- which is likely the case. But the implementation might still be a surprise.

go/processor/geth/processor.go
```diff
@@ -318,13 +322,13 @@ func (s *stateDB) Empty(address common.Address) bool {
...
 }
 
 func (s *stateDB) SlotInAccessList(address common.Address, slot common.Hash) (addressOk bool, slotOk bool) {
-	//lint:ignore SA1019 deprecated functions to be migrated
-	return s.context.IsSlotInAccessList(tosca.Address(address), tosca.Key(slot))
+	addressOk = bool(s.context.AccessAccount(tosca.Address(address)))
+	slotOk = bool(s.context.AccessStorage(tosca.Address(address), tosca.Key(slot)))
```
HerbertJordan, line 330: Same here, this implementation has unexpected side-effects.

I guess it is fine, since the `stateDB` is only intended for a single purpose, but one might have to explain this in a comment.

go/processor/geth/state_contract.go
HerbertJordan: Dependencies should be resolved on a module level, not a repository level. Are there actual dependencies between the module that contains the state contract and Tosca?

go/processor/geth/processor.go
HerbertJordan: Ok, but I think we should still replace the "ftm" part with "sonic". We do not aim for Fantom compatibility any more with Tosca.

go/processor/geth/processor.go
HerbertJordan: "new" is not a verb, but close enough ...

go/processor/geth/processor.go
HerbertJordan: The `stateDB` interface does not provide the typical error handling support where each function would return an error. This would make the handling of errors very complicated while running an EVM contract.

Instead, the idea is that errors are collected internally and operations are skipped if something fails. At the end of a transaction it is checked whether any errors have been encountered, and if so, the transaction is reverted. However, this error-check function is not part of the `StateDB` interface provided to the processor.

If we want the same error handling guarantees, we would have to do the same here. Have an error field in the `stateDB` and collect errors that are checked at the end of the processor run. Or, since it is a reference implementation that should not run in production and only needs to fit a single use case, a panic might be the way to go.

Ignoring it may be an option as well, but it pushes the responsibility to check that the balance is high enough to the caller.

go/processor/geth/processor.go
HerbertJordan: I do not think it is that easy. If the implementation would simply switch the order of the calls [here](https://github.com/Fantom-foundation/go-ethereum-sonic/blob/2400937cc3b1c370ea8735d7af86eac621e4eb70/core/vm/instructions.go#L807-L808) by first performing the self-destruct and then the AddBalance, the remaining tokens would go to the wrong account.

It is actually not the geth adapter that this is depending on, but the implementation of the self-destruct operation in geth.

go/processor/geth/processor.go
HerbertJordan: So factoring out synergies between the geth processor and interpreter is going to be implemented in a follow-up PR soon?

HerbertJordan (dismissed):

go/processor/geth/processor.go
simonlechner: Yes this will be my next PR, I didn't want to include code changes unrelated to the processor to this already pretty big PR.

go/processor/geth/processor.go
simonlechner: There should not be a call to this function outside of the `transfer` function in the `blockContext` and the `state contract`. Both perform a balance check before. I would ignore the error here, as the `context` implementation used in production also wont panic or set an error, it is the processors responsibility to ensure the balance is sufficient.

go/processor/geth/processor.go
simonlechner: I added comments to both functions.

go/processor/geth/processor.go
simonlechner: For all current implementations, these calls should be performed in the right order. I adapted the comment again.

go/processor/geth/state_contract.go
HerbertJordan: Alright, let's keep this as a future clean-up task.

go/processor/geth/processor.go
HerbertJordan: This dependency on the order of the calls should be documented -- and ideally a violation would be detected by at least one integration test. I assume the latter is the case.

go/processor/geth/processor.go
```diff
@@ -0,0 +1,405 @@
...
+	"github.com/ethereum/go-ethereum/core/vm"
+	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/trie/utils"
+)
+
+func init() {
+	tosca.RegisterProcessorFactory("geth-sonic", fantomProcessor)
+}
+
+func fantomProcessor(interpreter tosca.Interpreter) tosca.Processor {
```
HerbertJordan, line 35: nit:
```suggestion
func sonicProcessor(interpreter tosca.Interpreter) tosca.Processor {
```

HerbertJordan (approved):

go/processor/geth/processor.go
simonlechner: The comment documents the required call order.

PR #18, opened 2025-01-23
Combine geth to tosca state dbs

simonlechner:
The state db used for the geth-interpreter and geth-processor are identical, this PR merges them into a single file.

### DIFF
--- a/go/tosca_adapter/state_db.go
+++ b/go/tosca_adapter/state_db.go
@@ -1,0 +1,254 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tosca_adapter
+
+import (
+	"github.com/0xsoniclabs/tosca/go/tosca"
+	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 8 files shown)


go/tosca_adapter/state_db.go
```diff
@@ -0,0 +1,254 @@
...
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package tosca_adapter
```
HerbertJordan, line 11: Since this implementation is bridging the gap to the the geth interfaces I would suggest to include this file in the already existing `geth_adapter` package.

HerbertJordan (approved):
Looks good to me, thanks for the cleanup.

I would argue, though, that the `StateDB` adapter would fit well into the `geth_adapter` package. This package's role is to provide adapters facilitating the interaction between geth and Tosca -- independently of the direction.

LuisPH3 (dismissed):

go/tosca_adapter/state_db.go
simonlechner: I originally thought so too, but there has been a circular dependency in the [adapter_test](https://github.com/0xsoniclabs/tosca/blob/433748573a7737700d80ac48627e778fbbc4e5bd/go/geth_adapter/adapter_test.go#L30). I don't think this import is required, so I removed it and moved the state db to the geth adapter.

HerbertJordan (approved):

PR #19, opened 2025-01-29
Add gas computation to processor

simonlechner:
The price for one unit of gas is transaction specific, therefore the calculation of the gas price should be performed in the processor. The `tosca.Transaction` type is extended with a `GasFeeCap` and `GasTipCap` field rather than a single `GasPrice`.

### DIFF
--- a/go/examples/example.go
+++ b/go/examples/example.go
@@ -78,7 +78,7 @@ func (e *Example) RunOn(interpreter tosca.Interpreter, argument int) (Result, er
 func (e *Example) RunOnProcessor(processor tosca.Processor, argument int,
 	transaction tosca.Transaction, transactionContext tosca.TransactionContext) (Result, error) {
 
-	blockParameters := tosca.BlockParameters{}
+	blockParameters := tosca.BlockParameters{GasLimit: tosca.Gas(100000000)}
 	transaction.Input = encodeArgument(e.function, argument)
 
 	receipt, err := processor.Run(blockParameters, transaction, transactionContext)
--- a/go/examples/example.go
+++ b/go/examples/example.go
@@ -78,7 +78,7 @@ func (e *Example) RunOn(interpreter tosca.Interpreter, argument int) (Result, er
 func (e *Example) RunOnProcessor(processor tosca.Processor, argument int,
 	transaction tosca.Transaction, transactionContext tosca.TransactionContext) (Result, error) {
 
-	blockParameters := tosca.BlockParameters{GasLimit: tosca.Gas(100000000)}
+	blockParameters := tosca.BlockParameters{}
 	transaction.Input = encodeArgument(e.function, argument)
 
 	receipt, err := processor.Run(blockParameters, transaction, transactionContext)
--- a/go/integration_test/processor/scenario.go
+++ b/go/integration_test/processor/scenario.go
@@ -35,6 +35,10 @@ type Scenario struct {
 func (s *Scenario) Run(t *testing.T, processor tosca.Processor) {
 
 	context := newScenarioContext(s.Before)
+	if s.Parameters.GasLimit == 0 {
+		// Provide a sufficient gas limit for the block
+		s.Parameters.GasLimit = 1000000
+	}
 	receipt, err := processor.Run(s.Parameters, s.Transaction, context)
 	if err != nil && s.OperaError == nil {
 		t.Fatalf("failed to run transaction: %v", err)
--- a/go/integration_test/processor/scenario.go
+++ b/go/integration_test/processor/scenario.go
@@ -35,10 +35,6 @@ type Scenario struct {
 func (s *Scenario) Run(t *testing.T, processor tosca.Processor) {
 
 	context := newScenarioContext(s.Before)
-	if s.Parameters.GasLimit == 0 {
-		// Provide a sufficient gas limit for the block
-		s.Parameters.GasLimit = 1000000
-	}
 	receipt, err := processor.Run(s.Parameters, s.Transaction, context)
 	if err != nil && s.OperaError == nil {
 		t.Fatalf("failed to run transaction: %v", err)
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -54,6 +54,7 @@ func (p *processor) Run(
 		Success: false,
 		GasUsed: transaction.GasLimit,
 	}
+	gasPrice := calculateGasPrice(blockParameters.BaseFee, transaction.GasFeeCap, transaction.GasTipCap)
 	gas := transaction.GasLimit
 
 	if nonceCheck(transaction.Nonce, context.GetNonce(transaction.Sender)) != nil {
@@ -64,7 +65,7 @@ func (p *processor) Run(
 		return tosca.Receipt{}, nil
 	}
 
-	if err := buyGas(transaction, context); err != nil {
+	if err := buyGas(transaction, context, gasPrice); err != nil {
 		return tosca.Receipt{}, nil
 	}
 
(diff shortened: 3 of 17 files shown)


go/tosca/processor.go
```diff
@@ -30,7 +30,8 @@ type Transaction struct {
 	Input      Data          // the input data for the transaction
 	Value      Value         // the amount of network currency to transfer to the recipient
 	GasLimit   Gas           // the maximum amount of gas that can be used by the transaction
-	GasPrice   Value         // the effective price of a unit of gas for this transaction
+	GasFeeCap  Value         // the amount of network currency the sender is willing to pay for gas
```
LuisPH3, line 33:
```suggestion
	GasFeeCap  Value         // the amount of network currency the sender is willing to pay for one gas unit
```

go/tosca/processor.go
```diff
@@ -30,7 +30,8 @@ type Transaction struct {
...
+	GasTipCap  Value         // the amount of network currency the sender is willing to pay as a tip
```
LuisPH3, line 34:
```suggestion
	GasTipCap  Value         // the amount of network currency the sender is willing to pay for one gas unit as a prority fee
```

go/processor/opera/processor.go
```diff
@@ -289,6 +291,11 @@ func (p *processor) Run(
 	}, nil
 }
 
+func calculateGasPrice(baseFee, gasFeeCap, gasTipCap tosca.Value) tosca.Value {
+	gasPrice := tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```
LuisPH3, line 295:
```suggestion
	return tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```

go/processor/geth/processor.go
```diff
@@ -95,6 +101,11 @@ func (p *Processor) Run(
 	}, nil
 }
 
+func calculateGasPrice(baseFee, gasFeeCap, gasTipCap tosca.Value) tosca.Value {
+	gasPrice := tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```
LuisPH3, line 105:
```suggestion
	return tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```

go/processor/floria/processor.go
```diff
@@ -129,6 +130,11 @@ func (p *processor) Run(
 	}, nil
 }
 
+func calculateGasPrice(baseFee, gasFeeCap, gasTipCap tosca.Value) tosca.Value {
+	gasPrice := tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```
LuisPH3, line 134:
```suggestion
	return tosca.Add(baseFee, tosca.Min(gasTipCap, tosca.Sub(gasFeeCap, baseFee)))
```

LuisPH3 (approved):
Just some nits.

go/processor/geth/processor.go
```diff
@@ -159,12 +170,13 @@ func blockParametersToChainConfig(blockParams tosca.BlockParameters) *params.Cha
...
 }
 
-func newEVMConfig(interpreter tosca.Interpreter, ethereumCompatible bool) vm.Config {
+func newEVMConfig(interpreter tosca.Interpreter, noBaseFee bool, ethereumCompatible bool) vm.Config {
 	config := vm.Config{
 		StatePrecompiles: map[common.Address]vm.PrecompiledStateContract{
 			stateContractAddress: PreCompiledContract{},
 		},
 		Interpreter: geth_adapter.NewGethInterpreterFactory(interpreter),
+		NoBaseFee:   noBaseFee,
```
HerbertJordan, line 179: I think this field has a different semantic than the way you are using it here.

If `NoBaseFee` is set, the geth processor is internally setting the base-fee to zero. The sole use case for this is to allow RPC calls to accept any gas limits. This should not be relevant in our use cases.

go/processor/geth/processor.go
```diff
@@ -47,19 +47,25 @@ func (p *Processor) Run(
...
 	stateDB := geth_adapter.NewStateDB(context)
 	chainConfig := blockParametersToChainConfig(blockParameters)
-	config := newEVMConfig(p.interpreter, p.ethereumCompatible)
+	config := newEVMConfig(p.interpreter, noBaseFee, p.ethereumCompatible)
 	evm := vm.NewEVM(blockContext, txContext, stateDB, chainConfig, config)
 
-	msg := transactionToMessage(transaction, blockParameters.BaseFee)
-	gasPool := new(core.GasPool).AddGas(uint64(transaction.GasLimit))
+	msg := transactionToMessage(transaction, gasPrice)
+	gasPool := new(core.GasPool).AddGas(uint64(blockParameters.GasLimit))
```
HerbertJordan, line 68: This could (and maybe even should) use the transaction's `GasLimit` not the block's `GasLimit`, right?

I assume this is the reason why in so many tests the GasLimit needed to be added to the `BlockParameters`? This seems to be unreleated to the change of the Transaction interface. If it is indeed needed, it should be is own PR.

HerbertJordan (changes requested):

go/processor/geth/processor.go
simonlechner: I removed the changes and will create another PR. I will inlcude an explanation why this block gas limit is required.

go/examples/example.go
```diff
@@ -78,7 +78,7 @@ func (e *Example) RunOn(interpreter tosca.Interpreter, argument int) (Result, er
 func (e *Example) RunOnProcessor(processor tosca.Processor, argument int,
 	transaction tosca.Transaction, transactionContext tosca.TransactionContext) (Result, error) {
 
-	blockParameters := tosca.BlockParameters{}
+	blockParameters := tosca.BlockParameters{GasLimit: tosca.Gas(100000000)}
```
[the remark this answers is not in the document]
simonlechner: I removed this change and will create a follow up PR.

go/integration_test/processor/scenario.go
```diff
@@ -35,6 +35,10 @@ type Scenario struct {
 func (s *Scenario) Run(t *testing.T, processor tosca.Processor) {
 
 	context := newScenarioContext(s.Before)
+	if s.Parameters.GasLimit == 0 {
+		// Provide a sufficient gas limit for the block
+		s.Parameters.GasLimit = 1000000
```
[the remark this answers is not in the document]
simonlechner: I removed this change and will create a follow up PR.

go/processor/geth/processor.go
simonlechner: Right, I removed it form the config.

HerbertJordan (approved):

go/processor/floria/processor.go
```diff
@@ -129,6 +133,13 @@ func (p *processor) Run(
 	}, nil
 }
 
+func calculateGasPrice(baseFee, gasFeeCap, gasTipCap tosca.Value) (tosca.Value, error) {
```
LuisPH3, line 136: shall we create a package `go/processor/utils` `common` `shared` `...`, ?
This code is tested, but the analogous functions in `geth` and `opera` are not.

LuisPH3: Another option would be to add a method `CalculateGasPrice(baseFee)` to the `tosca.Transaction` type

LuisPH3 (dismissed):

go/processor/floria/processor.go
simonlechner: I would like keep the `Transaction` type a simple object holding just data. 
I will remove the `opera` implementation in a follow up PR and the function in `geth` has coverage through the integration tests.

LuisPH3 (approved):
:+1:

PR #20, opened 2025-01-29
Geth processor: Add blob support

simonlechner:
This PR extends the `tosca.Transaction` with the `blobHashes` and `blobBaseFeeCap`, required to enable blob transaction support. These fields are converted and forwarded in the geth processor.


LuisPH3 (approved):
This seems sufficient to me.
I do not see any gas being cheked/substracted, therefore blob gas consumption must be done somewhere down the line.
This PR makes blob transaction data available to dowstream functionality.

HerbertJordan (approved):

PR #21, opened 2025-01-31
Extend transaction context

simonlechner:
The transaction context interface has been extended by two functions, `CreateAccount` and `HasEmptyStateRoot`. Both are required by the processor to manage account life times in create and selfdestruct instructions.

### DIFF
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -427,13 +427,9 @@ func (a *runContextAdapter) CreateAccount(addr tosca.Address) {
 	a.evm.StateDB.CreateContract(gc.Address(addr))
 }
 
-func (a *runContextAdapter) HasEmptyStateRoot(addr tosca.Address) bool {
-	codeHash := a.evm.StateDB.GetCodeHash(gc.Address(addr))
+func (a *runContextAdapter) HasEmptyStorage(addr tosca.Address) bool {
 	rootHash := a.evm.StateDB.GetStorageRoot(gc.Address(addr))
-
-	return a.evm.StateDB.GetNonce(gc.Address(addr)) == 0 &&
-		(codeHash == gc.Hash{} || codeHash == types.EmptyCodeHash) &&
-		(rootHash == gc.Hash{} || rootHash == types.EmptyRootHash)
+	return rootHash == gc.Hash{} || rootHash == types.EmptyRootHash
 }
 
 func (a *runContextAdapter) AccountExists(addr tosca.Address) bool {
--- a/go/geth_adapter/adapter.go
+++ b/go/geth_adapter/adapter.go
@@ -420,17 +420,24 @@ func debugCallEnd(result tosca.CallResult, reserr error) {
 	}
 }
 
-func (a *runContextAdapter) AccountExists(addr tosca.Address) bool {
-	return a.evm.StateDB.Exist(gc.Address(addr))
+func (a *runContextAdapter) CreateAccount(addr tosca.Address) {
+	if !a.evm.StateDB.Exist(gc.Address(addr)) {
+		a.evm.StateDB.CreateAccount(gc.Address(addr))
+	}
+	a.evm.StateDB.CreateContract(gc.Address(addr))
 }
 
-func (a *runContextAdapter) CreateAccount(addr tosca.Address, code tosca.Code) bool {
-	if a.AccountExists(addr) {
-		return false
-	}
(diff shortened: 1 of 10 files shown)


go/tosca/interpreter.go
```diff
@@ -86,6 +86,8 @@ type RunContext interface {
 type TransactionContext interface {
 	WorldState
 
+	CreateAccount(Address)
```
LuisPH3, line 89: no error? 
what happens when the address given is a reserved (zero, or precompiled contract, or 0xdEaD) ? 
I am not sure what should happen, do you know if the contract execution can carry on when this happens?

HerbertJordan (changes requested):

go/tosca/interpreter.go
simonlechner: It is the processors job to make sure that this function is only called if it can be called.

go/tosca/interpreter.go
```diff
@@ -98,6 +100,10 @@ type TransactionContext interface {
 	EmitLog(Log)
 	GetLogs() []Log
 
+	// HasEmptyStateRoot returns whether the account has a 0 nonce,
+	// empty storage and empty code.
```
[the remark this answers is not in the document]
simonlechner: Updated.

go/geth_adapter/adapter.go
```diff
@@ -420,17 +420,20 @@ func debugCallEnd(result tosca.CallResult, reserr error) {
...
-func (a *runContextAdapter) CreateAccount(addr tosca.Address, code tosca.Code) bool {
-	if a.AccountExists(addr) {
-		return false
-	}
-	a.evm.StateDB.CreateAccount(gc.Address(addr))
-	a.evm.StateDB.SetCode(gc.Address(addr), code)
-	return true
+func (a *runContextAdapter) HasEmptyStorage(addr tosca.Address) bool {
+	rootHash := a.evm.StateDB.GetStorageRoot(gc.Address(addr))
+	return rootHash == gc.Hash{} || rootHash == types.EmptyRootHash
```
LuisPH3, line 432: - This may need of a documentation string.
- There is something I still do not understand: As we saw on Friday, the hash field could be dirty. How would the DB behave if a previously empty storage gets one key added in a previous TX? Would the hash remain zeroed ? .

LuisPH3 (dismissed):

go/geth_adapter/adapter.go
simonlechner: I added a documentation comment.
Dirty field should be handled by the database, I don't think the Processor is the right place to ensure hashes are updated.

LuisPH3: Me neither, I am quite confused about the Dirty thing. IMO a transaction must do an atomic modification of the database, and it is the database responsibility to return coherent values and not dirties after a transaction is finalized. This should be the definition of Transaction.

LuisPH3 (approved):

go/tosca/interpreter.go
HerbertJordan, line 1: These two new functions should be part of the `WorldState` interface, not the `TransactionContext`.

go/geth_adapter/adapter_test.go
```diff
@@ -278,8 +277,10 @@ func TestRunContextAdapter_AccountOperations(t *testing.T) {
 
 	stateDb.EXPECT().Exist(gc.Address(address)).Return(false)
 	stateDb.EXPECT().CreateAccount(gc.Address(address))
-	stateDb.EXPECT().SetCode(gc.Address(address), code)
-	created := adapter.CreateAccount(address, code)
+	stateDb.EXPECT().CreateContract(gc.Address(address))
+	stateDb.EXPECT().Exist(gc.Address(address)).Return(true)
+	adapter.CreateAccount(address)
+	created := adapter.AccountExists(address)
```
HerbertJordan, line 283: Isn't this simply checking whether you set up the assertion in line 281 correctly?

HerbertJordan (changes requested):
I think the new functions should be part of the `WorldState` interface, handling state information, not the `TransactionContext`, handling snapshots, logs and other temporary concepts.

go/geth_adapter/adapter_test.go
simonlechner: I wanted to ensure that both `CreateAccount` and `CreateContract` is called. I removed the exists check and added a comment.

HerbertJordan (approved):

PR #22, opened 2025-01-31
Set block gas limit

simonlechner:
Geth's gas pool is used per block, it ensures the upper bound for gas in a block is not exceeded. 
The processor checks the gas limit in the beginning of the transaction when the gas is bought.

### DIFF
--- a/go/processor/geth/processor_test.go
+++ b/go/processor/geth/processor_test.go
@@ -11,8 +11,11 @@
 package geth_processor
 
 import (
+	"errors"
 	"math/big"
 	"testing"
+
+	"github.com/ethereum/go-ethereum/core"
 
 	"github.com/0xsoniclabs/tosca/go/tosca"
 	"go.uber.org/mock/gomock"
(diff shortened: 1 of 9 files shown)


HerbertJordan:
I was under the impression that the task of the processor is to manage the execution of a individual transactions. I thus think that the enforcement of the block-gas limit is beyond the scope of what a processor is supposed to do.

Are there any reasons that made it mandatory that the enforcement of the block-gas limit is part of the processor?

HerbertJordan (commented):
I was under the impression that the task of the processor is to manage the execution of a individual transactions. I thus think that the enforcement of the block-gas limit is beyond the scope of what a processor is supposed to do.

Are there any reasons that made it mandatory that the enforcement of the block-gas limit is part of the processor?

go/processor/geth/processor_test.go
```diff
@@ -88,3 +91,25 @@ func TestGethProcessor_ConfigAddsStateContract(t *testing.T) {
 		t.Errorf("state contract not added to config")
 	}
 }
+
+func TestGethProcessor_BlockGasLimitIsEnforces(t *testing.T) {
```
LuisPH3, line 95:
```
func TestGethProcessor_BlockGasLimitIsEnforced(t *testing.T) {
```

LuisPH3 (dismissed):

HerbertJordan (commented):
I would like to understand why this is needed. Can you share a test case that fails without this change and passes with this change?

simonlechner:
> I would like to understand why this is needed. Can you share a test case that fails without this change and passes with this change?

I just added the ethereum compatible geth version to aida, checkout `simon/geth-ethereum` to run the ethereum tests. The branch uses the current main version of tosca. 
Here is an example input that fails without the changes of this PR.
`go run ./cmd/aida-vm-sdb/ ethtest --validate --evm-impl geth-eth --vm-impl lfvm ./path/to/tests/GeneralStateTests/stEIP1559/lowGasLimit.jso`

LuisPH3 (approved):

simonlechner:
The transaction processor should work on a transaction level and not include checks on a block level.

PR #23, opened 2025-02-03
Add geth processor registration

simonlechner:


go/processor/geth/processor.go
```diff
@@ -32,17 +32,21 @@ func init() {
...
-	}
+	return NewGethProcessor(interpreter, false)
 }
 
 type Processor struct {
 	interpreter        tosca.Interpreter
 	ethereumCompatible bool
 }
 
+func NewGethProcessor(interpreter tosca.Interpreter, ethereumCompatible bool) *Processor {
```
LuisPH3, line 43: nit: this function may not add much value. 
If you construct the struct directly, the arguments are named and the code may read better:
```
processor := Processor{
    interpreter: interpreter,
    ethereumCompatible: true,
}
```

Maybe even making them public? after all, the members are the same as the arguments to the constructor function.

LuisPH3 (approved):
The change introduce a package constructing a parametrized geth-processor, which can be optionally imported by consumer packages to enable its use or not.

go/processor/geth/processor.go
HerbertJordan: This would also avoid the bool flag parameter that is hurting readability on the call side of this function.

It is not clear what the `true` means in `NewGethProcessor(interpreter, true)`.

go/processor/geth_eth/processor.go
```diff
@@ -0,0 +1,24 @@
...
+
+package geth_processor_eth
+
+import (
+	geth_processor "github.com/0xsoniclabs/tosca/go/processor/geth"
+	"github.com/0xsoniclabs/tosca/go/tosca"
+)
+
+func init() {
+	tosca.RegisterProcessorFactory("geth-eth", ethereumProcessor)
```
HerbertJordan, line 19: Please add a line of documentation describing the properties of the processor that is registered under the name `geth-eth`.

HerbertJordan (approved):

go/processor/geth/processor.go
simonlechner: I think classes should have private fields whenever possible, but the argument regarding readability is valid. I removed the `NewGethProcessor` function.

HerbertJordan (approved):

LuisPH3 (approved):

PR #27, opened 2025-02-11
Geth: reset return data

simonlechner:
A nightly CT [run](https://scala.fantom.network/job/Tosca/job/Conformance%20Tests%20Geth/24/) failed.
The CT uses its own `Bytes` type which uses a golang string internally, for our use case an empty string and nil should be handled the same way.

### DIFF
--- a/go/ct/common/bytes.go
+++ b/go/ct/common/bytes.go
@@ -43,6 +43,13 @@ func RandomBytesOfSize(rnd *rand.Rand, size int) Bytes {
 	return NewBytes(data)
 }
 
+func (b Bytes) Eq(other Bytes) bool {
+	if (b.data == "" || b.data == "0x") && (other.data == "" || other.data == "0x") {
+		return true
+	}
+	return b.data == other.data
+}
+
 func (b Bytes) ToBytes() []byte {
 	return []byte(b.data)
 }
--- a/go/ct/common/bytes_test.go
+++ b/go/ct/common/bytes_test.go
@@ -149,3 +149,54 @@ func TestBytes_Get(t *testing.T) {
 		t.Errorf("unexpected slice, got %v", got)
 	}
 }
+
+func TestBytes_EqHandlesEmptyAndNilTheSameWay(t *testing.T) {
+	tests := map[string]struct {
+		first  Bytes
+		second Bytes
+		equal  bool
+	}{
+		"both empty": {
+			first:  NewBytes([]byte{}),
+			second: NewBytes([]byte{}),
+			equal:  true,
+		},
+		"first empty": {
+			first:  NewBytes([]byte{}),
+			second: NewBytes([]byte{1}),
+			equal:  false,
+		},
+		"second empty": {
+			first:  NewBytes([]byte{1}),
+			second: NewBytes([]byte{}),
+			equal:  false,
+		},
+		"both nil": {
+			first:  Bytes{},
+			second: Bytes{},
+			equal:  true,
+		},
+		"first nil": {
+			first:  Bytes{},
+			second: NewBytes([]byte{}),
+			equal:  true,
+		},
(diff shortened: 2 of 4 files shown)


LuisPH3 (dismissed):
The proposed change solves the issue of the == operator making a distinction between nil and [].

Another solution would be to re-implement the equal comparison of the state to be unequivocal:
- doing so could lead to misuses of the ==  operator
- not doing this change does not prevent from doing it in the future, since the CT will certainly find this error again.

HerbertJordan (changes requested):
Please merge in #31 and re-run the regression tests. On my system this revealed an issue.

simonlechner:
The `Bytes` type is used for `CallData`, `LastCallReturnData` and `ReturnData` for all of these fields empty and nil should be treated as equal. Therefore the `Bytes` type has been extended with an equals method.

go/ct/common/bytes.go
```diff
@@ -43,6 +43,13 @@ func RandomBytesOfSize(rnd *rand.Rand, size int) Bytes {
 	return NewBytes(data)
 }
 
+func (b Bytes) Eq(other Bytes) bool {
+	if (b.data == "" || b.data == "0x") && (other.data == "" || other.data == "0x") {
```
LuisPH3, line 47: I have to say that I find the 0x cases a little odd. I presume that they can only be generated when calling constructing the type without calling the new method. 
I do not see a bigger issue by having them. They are just surprising to me.

LuisPH3 (approved):
The later changes improve on the original pr by implementing a comparison that handles the original issue in the type itself.
:+1:

go/ct/common/bytes.go
HerbertJordan: I find it more than odd ... this code seems to implicate that the byte array is stored in the string using a hex-encoding, which should not be the case.

How the `Byte` type is supposed to be used: remember, that in Go the `string` type and the `[]byte` type are almost the same. The main difference are 
- a string can not be `nil`, just empty
- a string is immutable
- a string has operators like `=`, `<`, `>` implemented, while `[]byte` does not
- a string can be used as a key for a map

However, there is no restriction what is stored in the string. Anything that is stored in a `[]byte` can be stored in a `string`. And this is what is used here. The `Byte` type is an immutable `[]byte` equivalent that can be equally compared.

A `Byte` with the `string` "0x" is thus supposed to be equivalent with `[]byte{'0','x'}`, and not with the empty byte slice.

Also, by adding an `Eq` function, the main feature of having `=` support is abandoned.

HerbertJordan (changes requested):
About this:
> The CT uses its own Bytes type which uses a golang string internally, for our use case an empty string and nil should be handled the same way.

This is confusing, since `Bytes` can never be `nil`. It can only be empty.

The test case causing this problem seems to be the result of several conversion errors. The return data in the test input is "3078" which read in octet is '30' '78' = '0' 'x'. What has probably happend is that somewhere a `Byte` type is printed with the `%x` verb instead of the `%v` verb, that should be used. When using `%x` the `Byte` is first converted into a string using its `String()` function and the resulting string is hex-encoded.

So I guess there was an empty `Byte` at some point somewhere, which was printed as `0x` and become `3078` in some dump. When being re-loaded it was interpreted as hex, and the `Byte` value of `[]byte{0x30,0x78}` was loaded, which is not the same as the empty `Byte`.

Your change "fixes" this problem by covering up for those conversion issues by defining that `[]byte{}` is the same as `[]byte{0x30,0x78}`.

simonlechner:
This problem has already been fixed in #5, the nightly run failed due to jenkins still pointing to fantom-foundation/Tosca. But this failure surfaced problems with the serialization and the to string conversion of the `Bytes` type which will be fixed in a follow up PR.

PR #28, opened 2025-02-11
EVMZERO: Fix issue in C++ coverage system

LuisPH3:
At some point during development, the required programs to compute C++ coverage and prepare reports were removed. 
This was not detected because cmake keeps a cache in the developer machine.

This PR re-introduces required infrastructure for c++ coverage. 
This issue was detected during Issues migration from the old organization.


simonlechner (dismissed):
Good catch, thank you!

HerbertJordan (dismissed):
Thanks for identifying and fixing this!

simonlechner (approved):

PR #29, opened 2025-02-11
Update geth version to 1.15.0

simonlechner:
This PR upgrades the geth dependency to 1.15.0 and sets geth's newest supported revision to Prague.

The following tests pass
- [x] unit/integration tests
- [x] CT run using geth interpreter


go/geth_adapter/adapter.go
```diff
@@ -442,7 +444,7 @@ func (a *runContextAdapter) GetNonce(addr tosca.Address) uint64 {
 }
 
 func (a *runContextAdapter) SetNonce(addr tosca.Address, nonce uint64) {
-	a.evm.StateDB.SetNonce(gc.Address(addr), nonce)
+	a.evm.StateDB.SetNonce(gc.Address(addr), nonce, tracing.NonceChangeUnspecified)
```
LuisPH3, line 447: I guess this could be refined:
I have been reading the different values this constant can have [here](https://github.com/ethereum/go-ethereum/blob/cdb66c89d6f8de3afd5e8d72bb5a0e3f46c27362/core/tracing/hooks.go#L349)

It should not be very difficult to differentiate:
- nonce increment because of submitting a transaction (EOA)
- nonce in create (there are two enums we would have to differentiate)
- nonce in auth. 
- genesis
- revert

Shall we make a follow-up issue to deal with this detail so that the tracing is complete?

go/processor/opera/processor.go
```diff
@@ -620,7 +621,7 @@ func (preCompiledStateContract) Run(
 			return nil, 0, geth.ErrExecutionReverted
 		}
 
-		stateDB.SetNonce(acc, stateDB.GetNonce(acc)+value.Uint64())
+		stateDB.SetNonce(acc, stateDB.GetNonce(acc)+value.Uint64(), tracing.NonceChangeUnspecified)
```
LuisPH3, line 624: could this be `NonceChangeNewContract`?

go/interpreter/lfvm/gas_test.go
```diff
@@ -62,6 +62,9 @@ func TestGas_getDynamicCostsForSstore_exhaustive(t *testing.T) {
 
 	// Check that gas prices are computed correctly.
 	for _, revision := range tosca.GetAllKnownRevisions() {
+		if revision > newestSupportedRevision {
+			continue
+		}
```
LuisPH3, line 67: unexpected: what was the error?

go/integration_test/interpreter/verify_gas_test.go
```diff
@@ -126,6 +126,7 @@ func TestDynamicGas(t *testing.T) {
 							mockStateDB.EXPECT().GetCodeSize(gomock.Any()).AnyTimes()
 							// Also, they may create the account.
 							mockStateDB.EXPECT().SetNonce(gomock.Any(), uint64(1)).AnyTimes()
+							mockStateDB.EXPECT().GetCode(gomock.Any()).AnyTimes()
```
LuisPH3, line 129: someone is reading code during gas computation? 
this is interesting and foreshadowing.

HerbertJordan (changes requested):
This covers actually multiple changes that should be individual PRs:
- update of the geth dependency
- introduction of the Prague revision in the Tosca interfaces
- introduction of the Prague revision in the CT specification
- addition of support of Prague in the geth interpreter

Processing each one of those individually would make it way easier for reviewers and get some of those tasks faster merged then a PR covering all of it.

simonlechner:
This PR is split into several new ones, the changes are handled in #32 and following.

PR #30, opened 2025-02-11
EVMZERO: Enable fuzzing tests

LuisPH3:
During migration of old organization issues [issue 549](https://github.com/Fantom-foundation/Tosca/issues/549) was found.
The stated error cannot be reproduced anymore, and evmzero fuzzer (go-driven)  can be enabled.  This means that the fuzzer test will be treated as an unit test for each corpus (generated and stored)  during unit testing rounds.
Additionally it allows to fuzz the C++ vm.

- `go test -fuzz=FuzzEvmzero ./go/ct`
- `go test -fuzz=FuzzDifferentialEvmzeroVsGeth ./go/ct`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,6 @@ build/
 
 # Jenkinsfile linter config
 .groovylintrc.json
+
+# vscode c++ engine cache
+.cache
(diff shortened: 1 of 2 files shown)


HerbertJordan (approved):
Nice, thanks!

PR #31, opened 2025-02-12
Fix regression test to signal a failed execution if tests fail

HerbertJordan:
While reviewing other PRs I noticed that the regression team is not signaling a failing test to the test environment. This PR fixes that.


simonlechner (approved):
Looks good, thanks!

PR #32, opened 2025-02-12
Update geth dependency to 1.15

simonlechner:
Update of the geth dependency to version 1.15.0
Due to the removal of the `txContext` from the `NewEWM` function, it has to be set manually after the construction. 
Several functions in the stateDB interface changed due to the introduction of a separate tracing layer in geth, these functions now return the old value.


HerbertJordan (approved):
Looks good, thanks!

LuisPH3 (approved):
As discussed previously, uses of tracing.NonceChangeUnspecified could be improved.

To unblock this PR, this issue is to be addressed here: https://github.com/0xsoniclabs/tosca/issues/33

PR #37, opened 2025-02-17
Add CT support to check EIP-7702 semantics. 

LuisPH3:
[EIP-7702](https://eips.ethereum.org/EIPS/eip-7702) introduces a small change on dynamic costs for all four call interpreter instructions. 

This PR is organized by commits, visiting one commit at the time is recommended for review. 

**The EIP change:**
- The delegation-designator is a 23 byte slice which consists of the `0xef, 0x01, 0x00` prefix and a target account address (the delegate account).
- The delegation-designator is to be stored in the code storage of the account pointed by the second parameter of a call instruction, instead of a valid opcodes slice. 
-  Any call operation, finding a delegation-designator in the code storage of the address being called, needs to charge EIP-2929 gas costs based in the warm/cold status of the delegation address. 

This PR adds CT support to construct the states affected by these changes and the conditions to identify the inputs that exploit them. 

This PR increase the CT test count from 25507439914 to 228720519466  (~10 times). The change in state count influences CT execution time, and Unit test time. The later may not run anymore within the 10 minutes expectation.  

- [x] Run CT
- [x] Decide on viability of test count increment.
- [x] Improve unit test execution time.

### DIFF
--- a/go/ct/common/delegator_designator.go
+++ b/go/ct/common/delegator_designator.go
@@ -1,0 +1,43 @@
+// Copyright (c) 2024 Fantom Foundation
+//
+// Use of this software is governed by the Business Source License included
+// in the LICENSE file and at fantom.foundation/bsl11.
+//
+// Change Date: 2028-4-16
+//
+// On the date above, in accordance with the Business Source License, use of
+// this software will be governed by the GNU Lesser General Public License v3.
+
+package common
+
+import (
+	"bytes"
+
+	"github.com/0xsoniclabs/tosca/go/tosca"
+)
+
+// ParseDelegationDesignator returns the delegate from a code
+// segment containing a delegation designator, if any.
+// If the code segment does not contain a delegation designator,
+// second returned value is false, and the address shall be ignored.
+// see: https://eips.ethereum.org/EIPS/eip-7702
+func ParseDelegationDesignator(code Bytes) (tosca.Address, bool) {
+	raw := code.ToBytes()
+
+	if len(raw) < 23 {
+		return tosca.Address{}, false
+	}
+
+	if bytes.HasPrefix(raw, []byte{0xef, 0x01, 0x00}) {
+		var res tosca.Address
+		copy(res[:], raw[3:23])
+		return res, true
+	}
+	return tosca.Address{}, false
+}
+
+// NewDelegationDesignator creates a new delegation designator for the given address.
+// see: https://eips.ethereum.org/EIPS/eip-7702
+func NewDelegationDesignator(address tosca.Address) Bytes {
+	return NewBytes(append([]byte{0xef, 0x01, 0x00}, address[:]...))
+}
--- a/go/ct/common/delegator_designator.go
+++ b/go/ct/common/delegator_designator.go
@@ -24,7 +24,7 @@ import (
 func ParseDelegationDesignator(code Bytes) (tosca.Address, bool) {
 	raw := code.ToBytes()
 
-	if len(raw) < 23 {
+	if len(raw) != 23 {
 		return tosca.Address{}, false
 	}
 
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -444,7 +444,7 @@ func (a *delegationDesignatorConstraint) String() string {
 		return fmt.Sprintf("!isDelegated(%v)", a.address)
 	}
 	if a.delegateAccountStatus == tosca.WarmAccess {
-		return fmt.Sprintf("isDelegated(%v), warm(delegateOf(%v))", a.address, a.address)
+		return fmt.Sprintf("isDelegated(%v),warm(delegateOf(%v))", a.address, a.address)
 	}
-	return fmt.Sprintf("isDelegated(%v), cold(delegateOf(%v))", a.address, a.address)
+	return fmt.Sprintf("isDelegated(%v),cold(delegateOf(%v))", a.address, a.address)
 }
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -149,8 +149,18 @@ func (g *AccountsGenerator) String() string {
 		parts = append(parts, fmt.Sprintf("balance(%v) ≤ %v", con.address, con.value.DecimalString()))
 	}
 
+	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
+		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	})
 	for _, con := range g.DelegationDesignator {
-		parts = append(parts, fmt.Sprintf("DelegationDesignator(%v) = %v", con.addressVariable, con.accessKind))
+		access := "none"
+		if con.enabled {
+			access = "warm"
+			if con.accessKind == tosca.ColdAccess {
+				access = "cold"
+			}
+		}
+		parts = append(parts, fmt.Sprintf("DelegationDesignator(%v) = %v", con.addressVariable, access))
 	}
 
 	return "{" + strings.Join(parts, ",") + "}"
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -28,7 +28,7 @@ type AccountsGenerator struct {
 	minBalance           []balanceConstraint
 	maxBalance           []balanceConstraint
 	warmCold             []warmColdConstraint
-	DelegationDesignator []DelegationDesignatorConstraint
+	delegationDesignator []delegationDesignatorConstraint
 }
 
 func NewAccountGenerator() *AccountsGenerator {
@@ -41,7 +41,7 @@ func (g *AccountsGenerator) Clone() *AccountsGenerator {
 		warmCold:             slices.Clone(g.warmCold),
 		minBalance:           slices.Clone(g.minBalance),
 		maxBalance:           slices.Clone(g.maxBalance),
-		DelegationDesignator: slices.Clone(g.DelegationDesignator),
+		delegationDesignator: slices.Clone(g.delegationDesignator),
 	}
 }
 
@@ -53,7 +53,7 @@ func (g *AccountsGenerator) Restore(other *AccountsGenerator) {
 	g.warmCold = slices.Clone(other.warmCold)
 	g.minBalance = slices.Clone(other.minBalance)
 	g.maxBalance = slices.Clone(other.maxBalance)
-	g.DelegationDesignator = slices.Clone(other.DelegationDesignator)
+	g.delegationDesignator = slices.Clone(other.delegationDesignator)
 }
 
 func (g *AccountsGenerator) BindToAddressOfEmptyAccount(address Variable) {
@@ -98,17 +98,17 @@ func (g *AccountsGenerator) BindCold(address Variable) {
 	}
 }
 
-func (g *AccountsGenerator) BindDelegationDesignator(address Variable, accessKind tosca.AccessStatus) {
-	v := DelegationDesignatorConstraint{address, true, accessKind}
-	if !slices.Contains(g.DelegationDesignator, v) {
-		g.DelegationDesignator = append(g.DelegationDesignator, v)
+func (g *AccountsGenerator) BindDelegationDesignator(address Variable, delegateAccountStatus tosca.AccessStatus) {
+	v := delegationDesignatorConstraint{address, true, delegateAccountStatus}
+	if !slices.Contains(g.delegationDesignator, v) {
+		g.delegationDesignator = append(g.delegationDesignator, v)
 	}
 }
 
 func (g *AccountsGenerator) BindNoDelegationDesignator(address Variable) {
-	v := DelegationDesignatorConstraint{addressVariable: address, enabled: false}
-	if !slices.Contains(g.DelegationDesignator, v) {
-		g.DelegationDesignator = append(g.DelegationDesignator, v)
+	v := delegationDesignatorConstraint{addressVariable: address, isDelegated: false}
+	if !slices.Contains(g.delegationDesignator, v) {
+		g.delegationDesignator = append(g.delegationDesignator, v)
 	}
 }
 
@@ -179,21 +179,21 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 	}
 
 	// Check for conflicts among delegation designator constraints.
-	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
-		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	sort.Slice(g.delegationDesignator, func(i, j int) bool {
+		return g.delegationDesignator[i].Less(&g.delegationDesignator[j])
 	})
-	for i := 0; i < len(g.DelegationDesignator)-1; i++ {
-		a, b := g.DelegationDesignator[i], g.DelegationDesignator[i+1]
+	for i := 0; i < len(g.delegationDesignator)-1; i++ {
+		a, b := g.delegationDesignator[i], g.delegationDesignator[i+1]
 		// if a delegation designator constraint is conflicting with another one
 		// by having the same target address but different access kind or enabled
 		if a.addressVariable == b.addressVariable &&
-			(a.enabled != b.enabled || a.accessKind != b.accessKind) {
+			(a.isDelegated != b.isDelegated || a.delegateAccountStatus != b.delegateAccountStatus) {
 			return nil, fmt.Errorf("%w, address %v conflicting delegation designator constraints", ErrUnsatisfiable, a.addressVariable)
 		}
 	}
 
 	// Check for conflicts among empty constraints and delegation designator constraints.
-	for _, ddCon := range g.DelegationDesignator {
+	for _, ddCon := range g.delegationDesignator {
 		if slices.ContainsFunc(g.emptyAccounts, func(c emptyConstraint) bool {
 			return c.address == ddCon.addressVariable && c.empty
 		}) {
@@ -296,12 +296,12 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 
 	// Add delegation designator constraints.
 	// see: https://eips.ethereum.org/EIPS/eip-7702
-	for _, con := range g.DelegationDesignator {
+	for _, con := range g.delegationDesignator {
 		calleeAddress := getBoundOrBindNewAddress(con.addressVariable)
-		if con.enabled {
+		if con.isDelegated {
 			addr := getUnusedAddress()
 			accountsBuilder.SetCode(calleeAddress, NewDelegationDesignator(addr))
-			if con.accessKind == tosca.WarmAccess {
+			if con.delegateAccountStatus == tosca.WarmAccess {
 				accountsBuilder.SetWarm(addr)
 			}
 		} else {
@@ -415,10 +415,10 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
 	return a.value.Lt(b.value)
 }
 
-type DelegationDesignatorConstraint struct {
-	addressVariable Variable
-	enabled         bool
-	accessKind      tosca.AccessStatus
+type delegationDesignatorConstraint struct {
+	addressVariable       Variable
+	isDelegated           bool
+	delegateAccountStatus tosca.AccessStatus
 }
 
 func (a *DelegationDesignatorConstraint) Less(b *DelegationDesignatorConstraint) bool {
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -98,7 +98,7 @@ func (g *AccountsGenerator) BindCold(address Variable) {
 	}
 }
 
-func (g *AccountsGenerator) BindDelegationDesignator(address Variable, delegateAccountStatus tosca.AccessStatus) {
+func (g *AccountsGenerator) BindToAddressOfDelegatedAccount(address Variable, delegateAccountStatus tosca.AccessStatus) {
 	v := delegationDesignatorConstraint{address, true, delegateAccountStatus}
 	if !slices.Contains(g.delegationDesignator, v) {
 		g.delegationDesignator = append(g.delegationDesignator, v)
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -160,13 +160,28 @@ func (g *AccountsGenerator) String() string {
 }
 
 func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, accountAddress tosca.Address) (*st.Accounts, error) {
+
+	// This closure checks whenever to variables
+	// are the same variable or are already bound to the same value
+	areVariablesClashing := func(a, b Variable) bool {
+		if a == b {
+			return true
+		}
+		if aValue, isBoundA := assignment[a]; isBoundA {
+			if bValue, isBoundB := assignment[b]; isBoundB {
+				return aValue == bValue
+			}
+		}
+		return false
+	}
+
 	// Check for conflicts among empty constraints.
 	sort.Slice(g.emptyAccounts, func(i, j int) bool {
 		return g.emptyAccounts[i].Less(&g.emptyAccounts[j])
 	})
 	for i := 0; i < len(g.emptyAccounts)-1; i++ {
 		a, b := g.emptyAccounts[i], g.emptyAccounts[i+1]
-		if a.address == b.address && a.empty != b.empty {
+		if areVariablesClashing(a.address, b.address) && a.empty != b.empty {
 			return nil, fmt.Errorf("%w, address %v conflicting empty constraints", ErrUnsatisfiable, a.address)
 		}
 	}
@@ -179,7 +194,7 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 		a, b := g.delegationDesignator[i], g.delegationDesignator[i+1]
 		// if a delegation designator constraint is conflicting with another one
 		// by having the same target address but different access kind or enabled
-		if a.addressVariable == b.addressVariable &&
+		if areVariablesClashing(a.addressVariable, b.addressVariable) &&
 			(a.isDelegated != b.isDelegated || a.delegateAccountStatus != b.delegateAccountStatus) {
 			return nil, fmt.Errorf("%w, address %v conflicting delegation designator constraints", ErrUnsatisfiable, a.addressVariable)
 		}
@@ -188,7 +203,7 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 	// Check for conflicts among empty constraints and delegation designator constraints.
 	for _, ddCon := range g.delegationDesignator {
 		if slices.ContainsFunc(g.emptyAccounts, func(c emptyConstraint) bool {
-			return c.address == ddCon.addressVariable && c.empty
+			return areVariablesClashing(c.address, ddCon.addressVariable) && c.empty
 		}) {
 			return nil, fmt.Errorf("%w, address %v conflicting delegation designator - empty constraints", ErrUnsatisfiable, ddCon.addressVariable)
 		}
@@ -198,7 +213,7 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 	sort.Slice(g.warmCold, func(i, j int) bool { return g.warmCold[i].Less(&g.warmCold[j]) })
 	for i := 0; i < len(g.warmCold)-1; i++ {
 		a, b := g.warmCold[i], g.warmCold[i+1]
-		if a.key == b.key && a.warm != b.warm {
+		if areVariablesClashing(a.key, b.key) && a.warm != b.warm {
 			return nil, fmt.Errorf("%w, address %v conflicting warm/cold constraints", ErrUnsatisfiable, a.key)
 		}
 	}
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -149,18 +149,11 @@ func (g *AccountsGenerator) String() string {
 		parts = append(parts, fmt.Sprintf("balance(%v) ≤ %v", con.address, con.value.DecimalString()))
 	}
 
-	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
-		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	sort.Slice(g.delegationDesignator, func(i, j int) bool {
+		return g.delegationDesignator[i].Less(&g.delegationDesignator[j])
 	})
-	for _, con := range g.DelegationDesignator {
-		access := "none"
-		if con.enabled {
-			access = "warm"
-			if con.accessKind == tosca.ColdAccess {
-				access = "cold"
-			}
-		}
-		parts = append(parts, fmt.Sprintf("DelegationDesignator(%v) = %v", con.addressVariable, access))
+	for _, con := range g.delegationDesignator {
+		parts = append(parts, con.String())
 	}
 
 	return "{" + strings.Join(parts, ",") + "}"
@@ -421,9 +414,22 @@ type delegationDesignatorConstraint struct {
 	delegateAccountStatus tosca.AccessStatus
 }
 
-func (a *DelegationDesignatorConstraint) Less(b *DelegationDesignatorConstraint) bool {
+func (a *delegationDesignatorConstraint) Less(b *delegationDesignatorConstraint) bool {
 	if a.addressVariable != b.addressVariable {
 		return a.addressVariable < b.addressVariable
 	}
-	return b.accessKind == tosca.ColdAccess && a.accessKind != b.accessKind
+	if a.isDelegated != b.isDelegated {
+		return !a.isDelegated && b.isDelegated
+	}
+	return bool(!a.delegateAccountStatus && b.delegateAccountStatus)
+}
+
+func (a *delegationDesignatorConstraint) String() string {
+	if !a.isDelegated {
+		return fmt.Sprintf("!isDelegated(%v)", a.addressVariable)
+	}
+	if a.delegateAccountStatus == tosca.WarmAccess {
+		return fmt.Sprintf("isDelegated(%v), warm(delegateOf(%v))", a.addressVariable, a.addressVariable)
+	}
+	return fmt.Sprintf("isDelegated(%v), cold(delegateOf(%v))", a.addressVariable, a.addressVariable)
 }
--- a/go/ct/gen/accounts.go
+++ b/go/ct/gen/accounts.go
@@ -24,10 +24,11 @@ import (
 )
 
 type AccountsGenerator struct {
-	emptyAccounts []emptyConstraint
-	minBalance    []balanceConstraint
-	maxBalance    []balanceConstraint
-	warmCold      []warmColdConstraint
+	emptyAccounts        []emptyConstraint
+	minBalance           []balanceConstraint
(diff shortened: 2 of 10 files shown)


LuisPH3:
Unit tests cannot run within 10 min.
`panic: test timed out after 10m0s`

go/ct/spc/specification.go
```diff
@@ -2559,8 +2568,28 @@ func callEffect(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode) {
...
+	targetCode := s.Accounts.GetCode(target.Bytes20be())
+	delegateAddress, isDelegate := ParseDelegationDesignator(targetCode)
+	if s.Revision >= tosca.R14_Prague && isDelegate {
+		if s.Accounts.IsWarm(delegateAddress) {
+			DelegationDesignatorAccessCost = 100
+		} else {
+			s.Accounts.MarkWarm(delegateAddress)
+			DelegationDesignatorAccessCost = 2600
+		}
+	}
```
HerbertJordan, line 2583: I would assume that if it is a delegate address, we would need to remember the delegate address and somehow use it in the call we are creating below. Maybe as the `codeAddress` parameter?

LuisPH3: not quite.
There are 3 addresses involved when executing a CALL* opcode:
- A, the caller (sponsor)
- B, the callee (sponsored or authorizer)
- C, the delegate (location of the contract executed) 

An execution, in context of A, reaches a CALL* opcode with B as second argument (target). If the code in the account of B contains a D.Designator to C, then, the code executed by the call will be the code found at C.

This new indirection happens in the processor, but the dynamic cost is applied to the CALL* instruction and therefore the indirection shortly peak at by the interpreter. 

The address at `codeAddress` is B, which was already charged for cold/Warm status since EIP-2929

LuisPH3:
Last 2 commits reduce states generated:
```
before 228720519466
after   54437122858
```

go/ct/common/delegator_designator.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ParseDelegationDesignator returns the delegate from a code
+// segment containing a delegation designator, if any.
+// If the code segment does not contain a delegation designator,
+// second returned value is false, and the address shall be ignored.
+// see: https://eips.ethereum.org/EIPS/eip-7702
+func ParseDelegationDesignator(code Bytes) (tosca.Address, bool) {
+	raw := code.ToBytes()
+
+	if len(raw) < 23 {
```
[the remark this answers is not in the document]
LuisPH3: `!=` right!

go/ct/gen/accounts.go
```diff
@@ -147,6 +178,29 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
...
+	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
+		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	})
+	for i := 0; i < len(g.DelegationDesignator)-1; i++ {
+		a, b := g.DelegationDesignator[i], g.DelegationDesignator[i+1]
+		// if a delegation designator constraint is conflicting with another one
+		// by having the same target address but different access kind or enabled
+		if a.addressVariable == b.addressVariable &&
+			(a.enabled != b.enabled || a.accessKind != b.accessKind) {
+			return nil, fmt.Errorf("%w, address %v conflicting delegation designator constraints", ErrUnsatisfiable, a.addressVariable)
```
facuMH, line 191: would it not be better to append "unsatisfiable: conflict" instead of "unsatisfiable, conflict" ? the comma for me makes it look like listing a different error, not providing reason for the unsatisfiability

go/ct/gen/accounts.go
```diff
@@ -240,6 +294,22 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
...
+	// Add delegation designator constraints.
+	// see: https://eips.ethereum.org/EIPS/eip-7702
+	for _, con := range g.DelegationDesignator {
+		calleeAddress := getBoundOrBindNewAddress(con.addressVariable)
+		if con.enabled {
+			addr := getUnusedAddress()
+			accountsBuilder.SetCode(calleeAddress, NewDelegationDesignator(addr))
+			if con.accessKind == tosca.WarmAccess {
+				accountsBuilder.SetWarm(addr)
+			}
```
facuMH, line 306: should the else case be set as cold ?

go/ct/gen/accounts_test.go
```diff
@@ -96,6 +96,180 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
 	}
 }
 
+func TestAccountsGenerator_DelegationDesignatorCanBePrinted(t *testing.T) {
```
facuMH, line 99: maybe I am lacking some understanding, but I don't see the `DelegationDesignatorConstraint.enabled` condition being tested.

go/ct/rlz/conditions.go
```diff
@@ -1053,4 +1053,70 @@ func (c *hasNoBlobHash) String() string {
 	return fmt.Sprintf("%v does not have BlobHash", c.index.String())
 }
 
-////////////////////////////////////////////////////////////
+////////////////////////////////////////////////////////////////////////////////
+// EIP 7702
```
facuMH, line 1057: I think leaving this comment does not help much. I would rather see `Delegation Designation - https://eips.ethereum.org/EIPS/eip-7702`. If nothing else for consistency

go/ct/rlz/conditions.go
```diff
@@ -1053,4 +1053,70 @@ func (c *hasNoBlobHash) String() string {
...
+	delegateAddress, isDelegated := ParseDelegationDesignator(account.Code)
+	switch c.state {
+	case NoDelegationDesignation:
+		return !isDelegated, nil
+	case WarnDelegationDesignation:
+		return isDelegated && s.Accounts.IsWarm(delegateAddress), nil
+	case ColdDelegationDesignation:
+		return isDelegated && !s.Accounts.IsWarm(delegateAddress), nil
+	}
+	panic("unreachable")
```
facuMH, line 1091: is this debug code ? if not, could you maybe expand on the panic message please ?

go/ct/rlz/conditions.go
```diff
@@ -1053,4 +1053,70 @@ func (c *hasNoBlobHash) String() string {
...
+		case NoDelegationDesignation:
+			generator.BindNoDelegationDesignator(v)
+		case WarnDelegationDesignation:
+			generator.BindDelegationDesignator(v, tosca.WarmAccess)
+		case ColdDelegationDesignation:
+			generator.BindDelegationDesignator(v, tosca.ColdAccess)
+		}
+	}
+	return []TestValue{
+		NewTestValue(property, domain, c.state, restrict),
```
facuMH, line 1111: I believe this should return 3 test values, one with no delegation, one warm and one cold, no?

go/ct/rlz/conditions.go
```diff
@@ -1053,4 +1053,70 @@ func (c *hasNoBlobHash) String() string {
...
+		NewTestValue(property, domain, domain.SomethingNotEqual(c.state), restrict),
+	}
+}
+
+func (c *containsDelegationDesignation) Restrict(*gen.StateGenerator) {
+	panic("not needed")
```
facuMH, line 1117: if the condition's restrict is not needed why is the condition needed ?

go/ct/rlz/domains.go
```diff
@@ -422,3 +422,65 @@ func removeDuplicatesGeneric[T comparable](slice []T) []T {
...
+	return a == b
+}
+
+func (DelegationDesignatorDomain) Less(a DelegationDesignatorState, b DelegationDesignatorState) bool {
+	return a < b
+}
+
+func (DelegationDesignatorDomain) Predecessor(a DelegationDesignatorState) DelegationDesignatorState {
+	// this domain cannot be used with Lt() nor Gt()
+	panic("not ordered")
```
facuMH, line 465: for consistency, "not useful" is the message being used in other domains where some operator does not apply.

go/ct/rlz/domains.go
```diff
@@ -422,3 +422,65 @@ func removeDuplicatesGeneric[T comparable](slice []T) []T {
...
+		return a + 1
+	}
+	return NoDelegationDesignation
+}
+
+func (d DelegationDesignatorDomain) Samples(s DelegationDesignatorState) []DelegationDesignatorState {
+	return []DelegationDesignatorState{s}
+}
+
+func (DelegationDesignatorDomain) SamplesForAll(samples []DelegationDesignatorState) []DelegationDesignatorState {
```
facuMH, line 484: samples for all should return all interesting values of the domain, in this case all 3 values, no ?

go/ct/rlz/domains.go
LuisPH3: I took me quite some time to understand why the function was not implemented, I would rather prefer to explain the reason.

go/ct/rlz/domains.go
LuisPH3: yes an not. 
Samples return all interesting values for the testing of the property, that is, one value satisfying and one not. 
this reduces the test count, based on the assumption that the "wrong" value is wrong, no matter which one it is.

go/ct/rlz/conditions.go
LuisPH3: This is not debug, it is unreachable.

go/ct/rlz/conditions.go
LuisPH3: Samples return all interesting values for the testing of the property, that is, one value satisfying and one not. 
this reduces the test count, based on the assumption that the "wrong" value is wrong, no matter which one it is.

go/ct/gen/accounts.go
```diff
@@ -147,6 +178,29 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
 		}
 	}
 
+	// Check for conflicts among delegation designator constraints.
+	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
+		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	})
+	for i := 0; i < len(g.DelegationDesignator)-1; i++ {
```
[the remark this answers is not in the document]
LuisPH3: it is a sorted list, just like the other checks.

go/ct/gen/accounts.go
LuisPH3: This follows the existing errors,
Separating errors with other punctuation is not a good practice, the reason is that errors will be concatenated, this error could be included in another error, and then you have : several times. Coma is grammatically correct.
Lists of errors are not that usual, and may include other syntax.

go/ct/gen/accounts.go
LuisPH3: yes, the accountBuilder has a list of warm addresses, any address not marked, remains cold.

go/ct/gen/accounts_test.go
LuisPH3: - BindNoDelegationDesignator has enabled:false
- gen.BindDelegationDesignator(v, X) has enabled:true

go/ct/rlz/domains.go
LuisPH3: right, good catch.
This was trimmed under the assumption that this scopes the test space. It does not change the number of cases.
That happens on the condition test generation.

go/ct/rlz/conditions.go
LuisPH3: Documented as future work here #48

go/ct/gen/accounts.go
```diff
@@ -95,6 +98,20 @@ func (g *AccountsGenerator) BindCold(address Variable) {
 	}
 }
 
+func (g *AccountsGenerator) BindDelegationDesignator(address Variable, accessKind tosca.AccessStatus) {
```
HerbertJordan, line 101: Using `accessKind` as synonym for `AccessStatus` which in turn is actually a type with the values `ColdAccess` and `WarmAccess` used to inform whether a conduced access was warm or cold looks like something that should be cleaned up.

The variable should be renamed to something like "accountStatus", which can then be cold or warm. The "access" is what might happen during the execution, but the constraint is about the state before the execution. Defining a dedicated enum in this package for this field might be another step to separate these two concepts.

go/ct/gen/accounts.go
```diff
@@ -344,3 +414,16 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
 	}
 	return a.value.Lt(b.value)
 }
+
+type DelegationDesignatorConstraint struct {
+	addressVariable Variable
+	enabled         bool
```
HerbertJordan, line 420: `enable` sounds like this constraint can be `enabled`, in the sense that it is considered, and `disabled`, in the sense that it is ignored. A better name might be `isDelegating` or something similar, where both values express a constraint to be fulfilled.

go/ct/gen/accounts.go
```diff
@@ -344,3 +414,16 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
...
+	accessKind      tosca.AccessStatus
```
HerbertJordan, line 421: nit: "kind" and "status" should be aligned to avoid confusion.

go/ct/gen/accounts.go
```diff
@@ -344,3 +414,16 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
 	}
 	return a.value.Lt(b.value)
 }
+
+type DelegationDesignatorConstraint struct {
+	addressVariable Variable
```
HerbertJordan, line 419: nit: just `address`? The `Variable` part is in the type ...

go/ct/gen/accounts.go
```diff
@@ -344,3 +414,16 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
...
+	enabled         bool
+	accessKind      tosca.AccessStatus
+}
+
+func (a *DelegationDesignatorConstraint) Less(b *DelegationDesignatorConstraint) bool {
+	if a.addressVariable != b.addressVariable {
+		return a.addressVariable < b.addressVariable
+	}
+	return b.accessKind == tosca.ColdAccess && a.accessKind != b.accessKind
+}
```
HerbertJordan, line 429: This function does not consider the `enabled` field.

go/ct/gen/accounts.go
```diff
@@ -132,6 +149,20 @@ func (g *AccountsGenerator) String() string {
...
+	})
+	for _, con := range g.DelegationDesignator {
+		access := "none"
+		if con.enabled {
+			access = "warm"
+			if con.accessKind == tosca.ColdAccess {
+				access = "cold"
+			}
+		}
+		parts = append(parts, fmt.Sprintf("DelegationDesignator(%v) = %v", con.addressVariable, access))
```
HerbertJordan, line 163: The resulting output should be improved. Right now it would produce one of these values:
```
DelegationDesignator(x) = none
DelegationDesignator(x) = warm
DelegationDesignator(x) = cold
```
which is not very intuitive. Howe about printing this instead:
```
!Delegates(x)
Delegates(x), warm(deligationTarget(x))
Delegates(x), cold(delegationTarget(x))
```
This would be more in line with what the rest of this function is doing. The multiple clauses can be added individually to the `parts` slice, and if you sort the `parts` instead of the constraints, they would get ordered together with the other `cold` and `warm` constraints in the output. Wdyt?

go/ct/gen/accounts.go
```diff
@@ -147,6 +178,29 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
...
 
+	// Check for conflicts among delegation designator constraints.
+	sort.Slice(g.DelegationDesignator, func(i, j int) bool {
+		return g.DelegationDesignator[i].Less(&g.DelegationDesignator[j])
+	})
+	for i := 0; i < len(g.DelegationDesignator)-1; i++ {
+		a, b := g.DelegationDesignator[i], g.DelegationDesignator[i+1]
+		// if a delegation designator constraint is conflicting with another one
+		// by having the same target address but different access kind or enabled
+		if a.addressVariable == b.addressVariable &&
```
HerbertJordan, line 189: Technically it is insufficient to just check for the same variable names. What you actually would have to check is whether the values assigned to those variables in the assignment are the same.

For instance, assume there are the following constraints:
```
Delegates($a), !Delegates($b)
```
In the current code, these could never cause a conflict, since `$a` and `$b` are distinct variables. However, for whatever reasons, the initial assignment could be
```
a -> Address(1), b -> Address(1)
```
which would get you
```
Delegates(Address(1)), !Delegates(Address(1))
```
which is unsatisfied. This, however, is not detected by this code.

You can see this in this unit test:
```
func TestDelegateConstraints_MissesConflictsIntroducedByAssignment(t *testing.T) {
	vA := Variable("a")
	vB := Variable("b")

	generator := NewAccountGenerator()
	generator.BindDelegationDesignator(vA, tosca.ColdAccess)
	generator.BindNoDelegationDesignator(vB)

	rnd := rand.New(0)
	assignment := Assignment{}
	_, err := generator.Generate(assignment, rnd, tosca.Address{})
	if err != nil {
		t.Fatalf("Unexpected error during account generation")
	}

	assignment[vA] = NewU256(1)
	assignment[vB] = NewU256(1)
	_, err = generator.Generate(assignment, rnd, tosca.Address{})
	if err == nil {
		t.Fatalf("expected %v to be unsatisfiable", generator)
	}
}
```

However, I do suspect that issues like this are hidden in many generators and situations like this are currently not showing up in our constraints. So I guess we can ignore this. However, technically this makes this constraint solver incorrect.

go/ct/gen/accounts.go
```diff
@@ -147,6 +178,29 @@ func (g *AccountsGenerator) Generate(assignment Assignment, rnd *rand.Rand, acco
...
+			(a.enabled != b.enabled || a.accessKind != b.accessKind) {
+			return nil, fmt.Errorf("%w, address %v conflicting delegation designator constraints", ErrUnsatisfiable, a.addressVariable)
+		}
+	}
+
+	// Check for conflicts among empty constraints and delegation designator constraints.
+	for _, ddCon := range g.DelegationDesignator {
+		if slices.ContainsFunc(g.emptyAccounts, func(c emptyConstraint) bool {
+			return c.address == ddCon.addressVariable && c.empty
```
HerbertJordan, line 198: Also in this case, what you would actually have to check is the values mapped to those variables, not the variable names.

go/ct/gen/accounts.go
HerbertJordan, line 101: nit: grammar wise, this function should be named like "BindToAddressOfDelegatingAccount" since what the function is doing is adding a constraint that makes sure that the provided variable will get bound in the assignment to an address with a code that is a delegate designator.

HerbertJordan (commented):
I managed to review the generator part but have to stop now. I leave the comments I have so far, will continue the review tomorrow.

go/ct/gen/accounts.go
```diff
@@ -24,10 +24,11 @@ import (
...
 type AccountsGenerator struct {
-	emptyAccounts []emptyConstraint
-	minBalance    []balanceConstraint
-	maxBalance    []balanceConstraint
-	warmCold      []warmColdConstraint
+	emptyAccounts        []emptyConstraint
+	minBalance           []balanceConstraint
+	maxBalance           []balanceConstraint
+	warmCold             []warmColdConstraint
+	DelegationDesignator []DelegationDesignatorConstraint
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/ababa7cda92fa2822046c2eae728496ab2b8fe14

go/ct/gen/accounts.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/0d23b221d30a8e825695671d4bf4935ca1b29483

go/ct/gen/accounts.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/116027bc977b7a0d660c71ac0ba5fe058855ff06

go/ct/gen/accounts.go
LuisPH3: I implemented the format here: 116027bc977b7a0d660c71ac0ba5fe058855ff06
this is more in line with the existing constraints. 

Using the parts is a different story, given a variable for delegate, the warm/cold constraint would come out naturally, but with the current encapsulation is a little finicky to do it by parts.

go/ct/gen/accounts.go
```diff
@@ -344,3 +414,16 @@ func (a *balanceConstraint) Less(b *balanceConstraint) bool {
 	}
 	return a.value.Lt(b.value)
 }
+
+type DelegationDesignatorConstraint struct {
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/ababa7cda92fa2822046c2eae728496ab2b8fe14

go/ct/gen/accounts.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/ababa7cda92fa2822046c2eae728496ab2b8fe14

go/ct/gen/accounts.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/ababa7cda92fa2822046c2eae728496ab2b8fe14

go/ct/gen/accounts.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/f03a2db17d4eb9ab0dc0d3ed1cbc8ab53a2713fa

go/ct/gen/accounts.go
LuisPH3: This makes total sense to me, I was just following the existing code. 
I added the check here: https://github.com/0xsoniclabs/tosca/pull/37/commits/c6d48c143033e63a8112583ee55d13acf37f872d

Nevertheless this feels insuficient to me and the tests should be updated. I createad a follow up ticket because this falls  out of the scope of this PR: #52

go/ct/gen/accounts.go
LuisPH3: Idem. https://github.com/0xsoniclabs/tosca/pull/37/commits/c6d48c143033e63a8112583ee55d13acf37f872d

go/ct/gen/state.go
```diff
@@ -321,6 +321,19 @@ func (g *StateGenerator) IsAbsentBlobHashIndex(variable Variable) {
...
 }
 
+// BindToAddressOfDelegatedAccount constraints a variable to be bound to the
+// address of an account that is delegated, where the delegate address have
+// a defined AccessStatus.
+func (g *StateGenerator) BindToAddressOfDelegatedAccount(address Variable, delegateAccountStatus tosca.AccessStatus) {
+	g.accountsGen.BindToAddressOfDelegatedAccount(address, delegateAccountStatus)
+}
+
+// BindToAddressOfNotDelegatedAccount constraints a variable to be bound
```
HerbertJordan, line 331: non-matching name of function in documentation

go/ct/gen/state.go
```diff
@@ -321,6 +321,19 @@ func (g *StateGenerator) IsAbsentBlobHashIndex(variable Variable) {
...
+// to the address of an account that is not delegated.
+func (g *StateGenerator) BindToNotDelegatedAccount(address Variable) {
```
HerbertJordan, line 333: Should be renamed as well.

go/ct/rlz/domains.go
```diff
@@ -422,3 +422,69 @@ func removeDuplicatesGeneric[T comparable](slice []T) []T {
...
+
+// DelegationDesignatorState represents the state of a delegation designator in
+// the code of an account.
+// This type works together with ConstraintDelegationDesignator condition to model
+// EIP-7702 behavior.  see https://eips.ethereum.org/EIPS/eip-7702
+type DelegationDesignatorState = uint64
+
+const (
+	NoDelegationDesignation DelegationDesignatorState = iota
+	WarnDelegationDesignation
```
HerbertJordan, line 437:
```suggestion
	WarmDelegationDesignation
```

go/ct/rlz/domains.go
```diff
@@ -422,3 +422,69 @@ func removeDuplicatesGeneric[T comparable](slice []T) []T {
...
+	ColdDelegationDesignation
+)
+
+func DelegationDesignatorName(dd DelegationDesignatorState) string {
+	switch dd {
+	case NoDelegationDesignation:
+		return "no_delegation_designation"
+	case WarnDelegationDesignation:
+		return "warn_delegation_designation"
```
HerbertJordan, line 446:
```suggestion
		return "warm_delegation_designation"
```

go/ct/spc/specification.go
```diff
@@ -2426,7 +2435,7 @@ func getRulesForAllCallTypes() []Rule {
 	return res
 }
 
-func getRulesForCall(op vm.OpCode, revision tosca.Revision, warm, zeroValue bool, opEffect func(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode), static bool) []Rule {
+func getRulesForCall(op vm.OpCode, revision tosca.Revision, warm, zeroValue bool, DelegationDesignator DelegationDesignatorState, opEffect func(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode), static bool) []Rule {
```
HerbertJordan, line 2438: nit: variable names should start with a lower case

go/ct/spc/specification.go
```diff
@@ -2412,11 +2412,20 @@ func getRulesForAllCallTypes() []Rule {
 			for _, warm := range []bool{true, false} {
 				for _, static := range []bool{true, false} {
 					for _, zeroValue := range []bool{true, false} {
-						effect := callEffect
-						if op == vm.CALL && static && !zeroValue {
-							effect = callFailEffect
+						for _, DelegationDesignator := range []DelegationDesignatorState{NoDelegationDesignation, WarnDelegationDesignation, ColdDelegationDesignation} {
```
HerbertJordan, line 2415: nit: variables names should start with a lower case

go/ct/rlz/domains.go
```diff
@@ -422,3 +422,69 @@ func removeDuplicatesGeneric[T comparable](slice []T) []T {
...
+// EIP-7702 behavior.  see https://eips.ethereum.org/EIPS/eip-7702
+type DelegationDesignatorState = uint64
+
+const (
+	NoDelegationDesignation DelegationDesignatorState = iota
+	WarnDelegationDesignation
+	ColdDelegationDesignation
+)
+
+func DelegationDesignatorName(dd DelegationDesignatorState) string {
```
HerbertJordan, line 441: nit: functions should start with a verb.

HerbertJordan, line 441: This should probably be implement as
```suggestion
func (dd DelegationDesignatorState) String() string {
```

go/ct/spc/specification.go
```diff
@@ -2559,8 +2576,28 @@ func callEffect(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode) {
 		valueToEmptyAccountCost = 25000
 	}
 
+	// Apply costs releated to delegate designator.
+	// see https://eips.ethereum.org/EIPS/eip-7702
+	DelegationDesignatorAccessCost := tosca.Gas(0)
```
HerbertJordan, line 2581: nit: variable names should start with a lower case

go/ct/spc/specification.go
```diff
@@ -2559,8 +2576,28 @@ func callEffect(s *st.State, addrAccessCost tosca.Gas, op vm.OpCode) {
...
+	targetCode := s.Accounts.GetCode(target.Bytes20be())
+	delegateAddress, isDelegate := ParseDelegationDesignator(targetCode)
+	if s.Revision >= tosca.R14_Prague && isDelegate {
+		if s.Accounts.IsWarm(delegateAddress) {
+			DelegationDesignatorAccessCost = 100
+		} else {
+			s.Accounts.MarkWarm(delegateAddress)
+			DelegationDesignatorAccessCost = 2600
+		}
+	}
```
HerbertJordan, line 2591: nit: limit scope ..
```suggestion
	if s.Revision >= tosca.R14_Prague {
		targetCode := s.Accounts.GetCode(target.Bytes20be())
		delegateAddress, isDelegate := ParseDelegationDesignator(targetCode)
		if isDelegate {
			if s.Accounts.IsWarm(delegateAddress) {
				DelegationDesignatorAccessCost = 100
			} else {
				s.Accounts.MarkWarm(delegateAddress)
				DelegationDesignatorAccessCost = 2600
			}
		}
	}
```

LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/2a125decdb577d975e6f85525b6e598e878971ed

go/ct/gen/accounts_test.go
```diff
@@ -96,6 +96,179 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
...
+			} else {
+				if !accounts.Exists(calleeAddress) {
+					t.Errorf("Expected address to be a delegation designator but it is not")
+				}
+				account := accounts.GetAccount(calleeAddress)
+				delegateAddress, isDesignator := ParseDelegationDesignator(account.Code)
+				if !isDesignator {
+					t.Errorf("Expected address to be a delegation designator but it is not")
+				} else if test.access !=
+					tosca.AccessStatus(accounts.IsWarm(delegateAddress)) {
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/2a125decdb577d975e6f85525b6e598e878971ed

go/ct/rlz/domains.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/90ca95831b05441005e8f0d5c1168f19f48b3b34

LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/90ca95831b05441005e8f0d5c1168f19f48b3b34

go/ct/rlz/domains.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/030d2436f6a25b0865e6912b5a4b6f62946589f2

go/ct/gen/accounts.go
```diff
@@ -95,6 +98,20 @@ func (g *AccountsGenerator) BindCold(address Variable) {
 	}
 }
 
+func (g *AccountsGenerator) BindToAddressOfDelegatedAccount(address Variable, delegateAccountStatus tosca.AccessStatus) {
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/3286d6b244f3dec4963fe66bf2ff7107f3d9d8ad

go/ct/gen/accounts.go
```diff
@@ -95,6 +98,20 @@ func (g *AccountsGenerator) BindCold(address Variable) {
...
+	v := delegationDesignatorConstraint{address, true, delegateAccountStatus}
+	if !slices.Contains(g.delegationDesignator, v) {
+		g.delegationDesignator = append(g.delegationDesignator, v)
+	}
+}
+
+func (g *AccountsGenerator) BindNoDelegationDesignator(address Variable) {
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/3286d6b244f3dec4963fe66bf2ff7107f3d9d8ad

go/ct/gen/state.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/3286d6b244f3dec4963fe66bf2ff7107f3d9d8ad

go/ct/gen/accounts_test.go
```diff
@@ -96,6 +96,179 @@ func TestAccountsGenerator_WarmColdConstraintsNoAssignment(t *testing.T) {
...
+			},
+			expected: "{isDelegated($v1), warm(delegateOf($v1))}",
+		},
+		"is sorted": {
+			setup: func(gen *AccountsGenerator, v Variable) {
+				v2 := Variable("v2")
+				gen.BindToAddressOfDelegatedAccount(v2, tosca.WarmAccess)
+				gen.BindNoDelegationDesignator(v)
+			},
+			expected: "{!isDelegated($v1),isDelegated($v2), warm(delegateOf($v2))}",
```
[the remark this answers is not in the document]
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/516ec703331e4839a5876b22d05e495838c4a1d1

go/ct/rlz/domains.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/030d2436f6a25b0865e6912b5a4b6f62946589f2

go/ct/gen/state.go
LuisPH3: https://github.com/0xsoniclabs/tosca/pull/37/commits/3286d6b244f3dec4963fe66bf2ff7107f3d9d8ad

go/ct/gen/accounts.go
```diff
@@ -132,26 +149,71 @@ func (g *AccountsGenerator) String() string {
...
+	// are the same variable or are already bound to the same value
+	areVariablesClashing := func(a, b Variable) bool {
+		if a == b {
+			return true
+		}
+		if aValue, isBoundA := assignment[a]; isBoundA {
+			if bValue, isBoundB := assignment[b]; isBoundB {
+				return aValue == bValue
+			}
+		}
```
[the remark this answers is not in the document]
LuisPH3: I understand the problem, nevertheless this change is not so trivially included in this PR, since we have to consider such newly created addresses for the used map, and there is some code that needs to be rewritten and tested. 

This is to be done here: https://github.com/0xsoniclabs/tosca/issues/52
Shall I undo the changes introduced?  the lambda?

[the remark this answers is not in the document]
HerbertJordan: No, I think the lambda is fine. Maybe add a TODO in the lambda before the if with the "isBoundX" checks that this needs an improvement and reference #52.

go/ct/rlz/conditions.go
```diff
@@ -1053,4 +1053,71 @@ func (c *hasNoBlobHash) String() string {
...
+}
+
+func (c *containsDelegationDesignation) String() string {
+	return fmt.Sprintf("%v containsDelegationDesignation", c.address)
```
HerbertJordan, line 1122: This summary string is used for the `generator-info` and general printing of constraints for the end use. It should provide a hint which of the 3 situations - no delegation, delegation to cold address, delegation to warm address - the constraint is enforcing.

HerbertJordan (approved):
Looks good to me, thanks for building this!

go/ct/gen/accounts.go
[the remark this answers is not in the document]
LuisPH3: I extended comments in https://github.com/0xsoniclabs/tosca/issues/52#issuecomment-2673961413

The more I think about it, the less I see the need to extend it. The reason is that later uses of variables in the generator will use `getBoundOrBindNewAddress`, therefore unbound variables will end up having the same assignment and the current check is sufficient.

A different issue is whenever this is clean enough. I believe that this a pattern that affects all generators and all variables of any type. Maybe a holistic solution could be a good refactor.

facuMH (approved):

HerbertJordan (approved):

PR #40, opened 2025-02-18
evmrs: charge delegate resolution cost (prague update EIP-7702)

LorenzSchueler:
Charge the delegation resolution cost introduced in the Prague update.

Depends on #37
Closing 0xsoniclabs/sonic-admin#71


LuisPH3 (approved):
The proposed changes charge dynamic cost of delegation designation

simonlechner (dismissed):
As far as my rust skills go, it does not look wrong.

LuisPH3 (approved):

PR #50, opened 2025-02-19
CT: refactor json and string output

simonlechner:
The `Bytes`' type string and json output lead to some confusion in the past. The `String` and `MarshalJSON` methods have been unified.
The input state has been printed twice in case of a failure.
Clarification which value is from the CT and which from the interpreter under test has been added.

### DIFF
--- a/go/ct/common/bytes.go
+++ b/go/ct/common/bytes.go
@@ -14,6 +14,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"pgregory.net/rand"
 )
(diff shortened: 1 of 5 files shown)


LuisPH3 (commented):
This PR unifies serialization & toString to use hexadecimal notation (or empty string).

Can you please check that regressions can be parsed?  Otherwise, we need to migrate them.
```
Code": {
        "0x717d219e52c6082db7c9b91be004c062d8592b32": "5aa550006f38663a75a3eb5686d6c6f103dc7f83fe9d6fea5219290cdd9c147c5670ba3bc67b8202628e"
      }
```
Code does not seem to contain prefix 0x.

- [x] verfiy regression files

simonlechner:
> This PR unifies serialization & toString to use hexadecimal notation (or empty string).
> 
> Can you please check that regressions can be parsed? Otherwise, we need to migrate them.
> 
> ```
> Code": {
>         "0x717d219e52c6082db7c9b91be004c062d8592b32": "5aa550006f38663a75a3eb5686d6c6f103dc7f83fe9d6fea5219290cdd9c147c5670ba3bc67b8202628e"
>       }
> ```
> 
> Code does not seem to contain prefix 0x.
> 
>     * [ ]  verfiy regression files

All regression inputs got parsed in the CI, I checked selected regression input files that the code is parsed correctly.
The `0x` prefix is only trimmed if it is present.

LuisPH3 (approved):
:+1:

HerbertJordan (approved):
Thanks for fixing those inconsistencies!

PR #53, opened 2025-02-20
evmzero: Add eip-7702 delegation designation cost

simonlechner:
This PR adds the delegation designation billing to the `evmzero`. With these changes, the interpreter now fully supports the Prague revision.

- [x] Unit tests
- [x] updated CT run

### DIFF
--- a/cpp/vm/evmzero/interpreter.cc
+++ b/cpp/vm/evmzero/interpreter.cc
@@ -10,6 +10,7 @@
 
 #include "vm/evmzero/interpreter.h"
 
+#include <array>
 #include <bit>
 #include <cstdint>
 #include <cstdio>
@@ -1726,6 +1727,22 @@ struct CallImpl {
         }
       }
 
+      // EIP-7702 delegation designation warm/cold access cost
+      int64_t delegation_designation_cost = 0;
+      if (ctx.revision >= EVMC_PRAGUE) {
+        constexpr size_t kDelegationDesignationSize = 23;
+        std::array<uint8_t, kDelegationDesignationSize + 1> code;
+        size_t code_size = ctx.host->copy_code(account, 0, code.begin(), code.size());
+        if (code_size == kDelegationDesignationSize && code[0] == 0xef && code[1] == 0x01 && code[2] == 0x00) {
+          auto target = reinterpret_cast<evmc_address*>(&(code[3]));
+          if (ctx.host->access_account(*target) == EVMC_ACCESS_WARM) {
+            delegation_designation_cost = 100;
+          } else {
+            delegation_designation_cost = 2600;
+          }
+        }
+      }
+
       int64_t positive_value_cost = has_value ? 9000 : 0;
       int64_t value_to_empty_account_cost = 0;
       if constexpr (Op != op::CALLCODE) {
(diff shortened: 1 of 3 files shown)


cpp/vm/evmzero/interpreter_test.cc
```diff
@@ -14,9 +14,11 @@
 #include <gtest/gtest.h>
 
 #include <evmc/evmc.hpp>
+#include <iterator>
```
LuisPH3, line 17: Headers hygiene. It seems that in this project system headers are included first, after local header (the one with "") 
This is DPS syntax and I have learned to do it all the way round to reduce header bloat (system last). 
Anyway, it seems that this is still using DPS policies.

cpp/vm/evmzero/interpreter.cc
```diff
@@ -10,7 +10,9 @@
 
 #include "vm/evmzero/interpreter.h"
 
+#include <array>
 #include <bit>
+#include <cstddef>
```
LuisPH3, line 15: why is this header needed? 
I think i may be possible to remove it.

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1726,6 +1728,22 @@ struct CallImpl {
...
       }
 
+      // EIP-7702 delegation designation warm/cold access cost
+      int64_t delegation_designation_cost = 0;
+      if (ctx.revision >= EVMC_PRAGUE) {
+        constexpr size_t kDelegationDesignationSize = 23;
+        std::array<uint8_t, kDelegationDesignationSize + 1> code;
+        size_t code_size = ctx.host->copy_code(account, 0, code.begin(), code.size());
+        if (code_size == kDelegationDesignationSize && code[0] == 0xef && code[1] == 0x01 && code[2] == 0x00) {
+          auto target = reinterpret_cast<evmc_address*>(&(code[3]));
```
LuisPH3, line 1738: Note: this is an optimization to avoid one code copy. 
It is good that we have a unit test, but the ideal complement for this would be to run the unit tests using address-sanitizer.

simonlechner: The CI unit test run with an address-sanitzer, I just debugged an asan error.

cpp/vm/evmzero/interpreter_test.cc
simonlechner: Since it is done that way in all files, I would prefer to have all changed in a separate PR or keep it as it is.

LuisPH3 (approved):

cpp/vm/evmzero/interpreter.cc
```diff
@@ -1726,6 +1727,22 @@ struct CallImpl {
         }
       }
 
+      // EIP-7702 delegation designation warm/cold access cost
+      int64_t delegation_designation_cost = 0;
+      if (ctx.revision >= EVMC_PRAGUE) {
+        constexpr size_t kDelegationDesignationSize = 23;
+        std::array<uint8_t, kDelegationDesignationSize + 1> code;
```
HerbertJordan, line 1734: Why the `+ 1`?

HerbertJordan (changes requested):
Looks good, I would just like to see the cold case also covered in the unit test.

cpp/vm/evmzero/interpreter.cc
simonlechner: It is only a delegation designation if the code length is 23. It is not necessary to copy the whole code but it has to be ensured that longer codes are not cut to the delegation size and treated as a delegation.

LuisPH3 (dismissed):

cpp/vm/evmzero/interpreter.cc
HerbertJordan: Nice, maybe add a line of comment there, this is important.

HerbertJordan (dismissed):

LuisPH3 (approved):

PR #66, opened 2025-04-10
evmrs: constrain lifetimes of references created from pointers

LorenzSchueler:
When creating references from pointers obtained via FFI, the lifetime is not constrained in any way when using `unsafe { &*ptr }` or [`slice_from_raw_parts`](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html).
This PR constraints the lifetime of those references by the lifetime of a token which is created at the beginning of the functions which get called via FFI.
This way, the references live only as long as the reference of the token - so only until the end of the scope, which is exactly what is expected of them.

### DIFF
--- a/rust/src/ffi/evmc_vm.rs
+++ b/rust/src/ffi/evmc_vm.rs
@@ -1,6 +1,6 @@
 use std::{
     ffi::{CStr, c_char},
-    panic, slice,
+    panic,
 };
 
 use evmc_vm::{
@@ -12,7 +12,12 @@ use evmc_vm::{
     },
 };
 
-use crate::evmc::EvmRs;
+use crate::{
+    evmc::EvmRs,
+    ffi::{
+        LifetimeToken, ref_from_ptr_scoped, ref_mut_from_ptr_scoped, slice_from_raw_parts_scoped,
+    },
+};
 
 static EVM_RS_NAME: &CStr = c"evmrs";
 static EVM_RS_VERSION: &CStr = c"0.1.0";
@@ -30,6 +35,8 @@ extern "C" fn __evmc_set_option(
     key: *const c_char,
     value: *const c_char,
 ) -> evmc_set_option_result {
+    let token = LifetimeToken;
+
     assert!(!instance.is_null());
 
     if key.is_null() {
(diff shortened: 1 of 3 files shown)


LuisPH3 (commented):
@psalz @ihtefyw 
This PR contains very interesting properties, I invite you guys to add a review comment, describing what does this PR change and what do we get from it.

ihtefyw:
My take could be a bit simplistic but: in my understanding, we want everything to have a lifetime in order to let the borrow checker warn us about wrong usage of the type, and pointers from FFI cannot have lifetimes (I suppose because they need to be freed explicitly?).
To make the pointer usage more secure it allocates a local object and, whenever it needs to use the pointer, a reference with a lifetime bound to the local token is created. In this way, the reference cannot live more than the token and therefore the pointer becomes "local" to our function call. I suppose this works until we construct a reference every time a pointer is needed.
However, I don't know if there are additional reasons for having it. Is there any other practical reason for binding the pointer to a reference with a specific lifetime?

rust/src/ffi/evmc_vm.rs
```diff
@@ -118,22 +128,23 @@ extern "C" fn __evmc_execute(
 
     // SAFETY:
     // `message`` is not null. The caller must make sure it points to a valid `ExecutionMessage`.
```
psalz, line 130:
```suggestion
    // `message` is not null. The caller must make sure it points to a valid `ExecutionMessage`.
```

psalz (commented):
With my limited understanding of lifetimes, I interpret the change as follows: The lifetime of objects pointed to by raw pointers passed into FFIs is generally unknown. The only thing that can be said definitively is that they remain valid for the duration of the FFI call itself. When creating references / slices from raw pointers, their lifetime is initially unconstrained. To avoid user-after-free bugs, the new set of `_scoped` helper functions tie the lifetime of the dereferenced objects to a *token* that lives inside the scope of the function called via FFI, thereby preventing them from being moved outside the function.

Here's my question, though: Since the entry point is the FFI function, the only way we could "leak" the lifetime of objects in a way that might cause issues would be to either store them inside a static variable, or to move them into one of the other parameters passed into the function - correct?

LorenzSchueler:
@psalz your explanation is spot on.
And yes, we basically want to make sure that any data passed as a pointer into `__evmc_execute` is not used after the function returns. And the only way that could happen (at least what I can think of now) is either storing it in a static or storing it in one of the parameters, as you already pointed out. In particular, the `instance` field is a pointer to the `EvmRs` instance which holds caches, so that would be the most likely candidate to *mess this up*.
@ihtefyw I think you also have the idea right. One thing I would like to point out is that pointers in general have no lifetimes. This has nothing to do with whether they are passed via FFI or not. Therefore, when converting a pointer to a reference (which is obviously only possible in unsafe Rust), the compiler is free to infer any lifetime that fulfills the requirements - making it essentially unconstrained.

LuisPH3 (approved):

PR #70, opened 2025-04-10
lfvm: Check whether code size exceeds maximum

simonlechner:
The lfvm uses an uint16 PC and can therefore not work with codes greater than max uint16. The maximum size of codes that can be executed on chain are init codes which are limited to 49152 bytes by ethereum.  
An error is now returned if the code size exceeds 49152 opcodes/bytes.

### DIFF
--- a/go/interpreter/lfvm/lfvm.go
+++ b/go/interpreter/lfvm/lfvm.go
@@ -124,6 +124,10 @@ func (v *lfvm) Run(params tosca.Parameters) (tosca.Result, error) {
 		return tosca.Result{}, &tosca.ErrUnsupportedRevision{Revision: params.Revision}
 	}
 
+	if len(params.Code) > MaxInitCodeSize {
+		return tosca.Result{}, errCodeSizeExceeded
+	}
+
 	converted := v.converter.Convert(
 		params.Code,
 		params.CodeHash,
--- a/go/interpreter/lfvm/errors.go
+++ b/go/interpreter/lfvm/errors.go
@@ -24,4 +24,5 @@ const (
 	errMaxMemoryExpansionSize = tosca.ConstError("max memory expansion size exceeded")
 	errStackUnderflow         = tosca.ConstError("stack underflow")
 	errStackOverflow          = tosca.ConstError("stack overflow")
+	errCodeSizeExceeded       = tosca.ConstError("max code size exceeded")
 )
--- a/go/interpreter/lfvm/instructions.go
+++ b/go/interpreter/lfvm/instructions.go
@@ -18,6 +18,11 @@ import (
 	"github.com/holiman/uint256"
 )
 
+const (
+	MaxCodeSize     = 24576           // Maximum bytecode to permit for a contract
+	MaxInitCodeSize = 2 * MaxCodeSize // Maximum initcode to permit in a creation transaction and create instructions
+)
+
 func opStop() status {
 	return statusStopped
 }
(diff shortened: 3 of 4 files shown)


go/interpreter/lfvm/lfvm.go
```diff
@@ -124,6 +124,10 @@ func (v *lfvm) Run(params tosca.Parameters) (tosca.Result, error) {
 		return tosca.Result{}, &tosca.ErrUnsupportedRevision{Revision: params.Revision}
 	}
 
+	if len(params.Code) > MaxInitCodeSize {
+		return tosca.Result{}, errCodeSizeExceeded
+	}
```
HerbertJordan, line 129: I think this check should actually go into the converter, since it is the component imposing this limit. If it is checked inside the converter, no other potential user of the converter has to recall this limitation.

A thought: The actual limit of the converter is 64k; so it could accept everything up to that.

HerbertJordan (changes requested):
I think the code-size check should be part of the converter. Wdyt?

go/interpreter/lfvm/lfvm.go
simonlechner: Yes the check should be in the converter, I moved it there.
I set the max code size to the max init code size because I thought it does not make sense to simulate code that can not be executed on the chain. But if there are chances in the future, or the interpreter is used somewhere else it makes sense to define the actual limit. I changed it to maxUint16.

go/interpreter/lfvm/converter_test.go
```diff
@@ -12,16 +12,18 @@ package lfvm
 
 import (
 	"bytes"
+	goContext "context"
```
LuisPH3, line 15: delete, please `t.Context()` instead

(`goContext` is a horrible horrible name, the whole story that there is a standard library object called context still gives me the chills)

go/interpreter/lfvm/lfvm.go
```diff
@@ -124,10 +124,13 @@ func (v *lfvm) Run(params tosca.Parameters) (tosca.Result, error) {
 		return tosca.Result{}, &tosca.ErrUnsupportedRevision{Revision: params.Revision}
 	}
 
-	converted := v.converter.Convert(
+	converted, err := v.converter.Convert(
 		params.Code,
 		params.CodeHash,
 	)
+	if err != nil {
+		return tosca.Result{}, err
```
LuisPH3, line 132: nit: 
```suggestion
		return tosca.Result{}, fmt.Errorf("failed to convert code: %w", err)
```

go/interpreter/lfvm/ct.go
```diff
@@ -49,10 +49,13 @@ func (a *ctAdapter) StepN(state *st.State, numSteps int) (*st.State, error) {
 		return state, nil
 	}
 
-	converted := a.vm.converter.Convert(
+	converted, err := a.vm.converter.Convert(
 		params.Code,
 		params.CodeHash,
 	)
+	if err != nil {
+		return &st.State{}, err
```
LuisPH3, line 57:
```suggestion
		return &st.State{}, fmt.Errorf("failed to convert code: %w", err)
```

go/interpreter/lfvm/converter_test.go
```diff
@@ -147,20 +210,29 @@ func TestConverter_ConverterIsThreadSafe(t *testing.T) {
 	code := []byte{byte(vm.STOP)}
 	hash := tosca.Hash{byte(1)}
 
-	var wg sync.WaitGroup
-	wg.Add(NumGoroutines)
-	for i := 0; i < NumGoroutines; i++ {
-		go func(i int) {
-			defer wg.Done()
+	errs, _ := errgroup.WithContext(goContext.Background())
```
LuisPH3, line 213: use `t.Context()`

LuisPH3 (approved):

LuisPH3 (approved):

PR #78, opened 2025-04-16
evmrs: use empty containers instead of optional ones

LorenzSchueler:
This PR changes the type for the fields `input`, `code`, `output` and `last_call_return_data` from optional slices or optional boxed slices to slices or boxes slices.
Semantically, there is no difference between having no input or empty input (the same also holds for the other fields).
By making the fields non-optional, it is now sufficient to check if they are empty instead of checking if they are None or empty. This also removes branching overall.

### DIFF
--- a/rust/evmc-vm/src/lib.rs
+++ b/rust/evmc-vm/src/lib.rs
@@ -67,7 +67,7 @@ pub struct ExecutionResult {
     pub status_code: StatusCode,
     pub gas_left: i64,
     pub gas_refund: i64,
-    pub output: Option<Box<[u8]>>,
+    pub output: Box<[u8]>,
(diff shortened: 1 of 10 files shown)


rust/evmc-vm/src/lib.rs
```diff
@@ -409,45 +415,25 @@ impl From<ffi::evmc_step_result> for StepResult {
...
-            {
-                None
-            } else {
-                Some(unsafe {
-                    Box::from(slice::from_raw_parts(
-                        value.last_call_return_data as *mut _,
-                        value.last_call_return_data_size,
-                    ))
-                })
-            },
```
LuisPH3, line 450: this is one big improvement

PR #79, opened 2025-04-18
evmrs: make cache sizes configurable

LorenzSchueler:
This PR makes the size of the code analysis and hash cache configurable by ~~setting environment variables~~ calling `set_option`.
This requires the following changes:
- the caches are no longer stored in statics, but in the EvmRs instance.
- references to the caches have to be passed around
- the GetGenericStatic trait is removed because with the changes there is now only a single use (for the jumptable) as it is easier to just implement the functionality there directly.


LuisPH3 (changes requested):
Unfortunately, we do not use environment variables to tune internal behavior of libraries. 
The standard mechanism to pass information is to use configuration files or command line flags. Which are passed to internal tools via call arguments and variables.

rust/src/evmrs.rs
```diff
@@ -62,10 +62,25 @@ impl EvmcVm for EvmRs {
...
+                    return Err(SetOptionError::InvalidValue);
+                }
+            }
+            ("hash-cache-size", size) => {
+                if let Ok(size) = size.parse::<usize>() {
+                    self.hash_cache = HashCache::new(size);
+                } else {
+                    return Err(SetOptionError::InvalidValue);
+                }
+            }
```
LuisPH3, line 83: how can we test this?

LorenzSchueler: Added tests for it

LuisPH3 (approved):

PR #80, opened 2025-04-24
Geth adapter: Refactor isPrecompiledContract function

simonlechner:
The readOnly flag is passed to geth using the msb of the gas. In case of a call to a precompiled contract this can not be done. The function which checks if the address is a precompiled contract has not been updated for revisions newer than Shanghai.


LuisPH3 (commented):
I propose that we discuss these changes briefly before merging.

simonlechner:
Refactoring of the geth adapter is postponed until the Prague revision has been rolled out.

PR #81, opened 2025-05-21
Geth adapter: Refactor isprecompiled function

simonlechner:
The `isPrecompiledContract` function in the geth adapter is hard to read and error prone, this refactor uses geth's precompile definitions to check whether an address is a precompiled contract.


HerbertJordan (dismissed):
Nice cleanup, thanks!

facuMH (approved):

LuisPH3 (approved):
:+1:

PR #83, opened 2025-05-23
Geth adapter: Add unit tests

simonlechner:
This PR adds a few unit tests to the geth adapter focusing on the correct forwarding and handling of errors and values.
Note the changes in `go.mod` and `go.sum` files are due to the import of the `requires` package.

### DIFF
--- a/go/geth_adapter/adapter_test.go
+++ b/go/geth_adapter/adapter_test.go
@@ -607,6 +607,14 @@ func TestGethAdapter_CorruptValuesReturnErrors(t *testing.T) {
 }
 
 func TestGethAdapter_CallForwardsToTheRightKind(t *testing.T) {
+	sender := common.Address{0x42}
+	recipient := common.Address{0x43}
+	codeAddress := common.Address{0x44}
+	input := []byte{0x01, 0x02, 0x03}
+	gas := uint64(1000)
+	value := uint256.NewInt(100)
+	salt := uint256.NewInt(200)
+
 	any := gomock.Any()
 	tests := map[string]struct {
 		kind  tosca.CallKind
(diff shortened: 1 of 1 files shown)


facuMH (changes requested):
A couple of error test suggestions:
- `runContextAdapter.Call` with an unsupported call kind.
- `getInterpreterAdapter,Run` reverted is correctly propagated
- `getInterpreterAdapter,Run`  exceeds call depth

Not an error but also worth testing (perhaps in a follow-up):
- `getInterpreterAdapter,Run`  successful execution with negative refund reduces gas refund. 

Look below for some minor code comments.

go/geth_adapter/adapter_test.go
```diff
@@ -548,6 +538,180 @@ func TestRunContextAdapter_Run(t *testing.T) {
...
+		chainID     *big.Int
+		blobBaseFee *big.Int
+		gasPrice    *big.Int
+		difficulty  *big.Int
+	}{
+		"revision": {
+			firstBlock: big.NewInt(1000),
+		},
+		"baseFee": {
+			baseFee: big.NewInt(-1),
```
HerbertJordan, line 554: Would a `big.Int` that is equal or larger than 2^256 be detected and correctly handled somewhere?

go/geth_adapter/adapter_test.go
```diff
@@ -548,6 +538,180 @@ func TestRunContextAdapter_Run(t *testing.T) {
...
+func TestGethAdapter_CallForwardsToTheRightKind(t *testing.T) {
+	any := gomock.Any()
+	tests := map[string]struct {
+		kind  tosca.CallKind
+		setup func(mock *MockCallContextInterceptor)
+	}{
+		"call": {
+			kind: tosca.Call,
+			setup: func(mock *MockCallContextInterceptor) {
+				mock.EXPECT().Call(any, any, any, any, any, any).Return(
```
HerbertJordan, line 618: fyi: if the expected return of the call should be all zero-values you can also omit the `.Return(...)`; but if you think this makes it clearer, feel free to keep it as it is. I am fine with both.

go/geth_adapter/adapter_test.go
```diff
@@ -548,6 +538,180 @@ func TestRunContextAdapter_Run(t *testing.T) {
...
+					nil, uint64(0), nil,
+				)
+			},
+		},
+		"delegateCall": {
+			kind: tosca.DelegateCall,
+			setup: func(mock *MockCallContextInterceptor) {
+				mock.EXPECT().DelegateCall(any, any, any, any, any).Return(
```
HerbertJordan, line 626: nit: This test might also be a good opportunity to check that the call parameters are forwarded correctly - or would this be an independent test?

HerbertJordan (dismissed):
The PR adds a range of unit tests verifying the proper forwarding of parameters from the geth-API to the Tosca API and vice versa.

The covered properties are a good addition. However, a few more properties should get covered. Among the things that should be added in follow-up PRs (unless already covered otherwise):
- the correct usage of `decodeReadOnlyFromGas`
- the correct usage of "refund shift"

go/geth_adapter/adapter_test.go
```diff
@@ -548,6 +538,180 @@ func TestRunContextAdapter_Run(t *testing.T) {
...
+	blockContext := geth.BlockContext{
+		BlockNumber: big.NewInt(24),
+	}
+	evm := geth.NewEVM(blockContext, nil, chainConfig, geth.Config{})
+	adapter := &runContextAdapter{evm: evm}
+	_, err := adapter.Call(tosca.Call, tosca.CallParameters{})
+	require.Error(t, err, "unsupported revision")
+}
+
+func TestGethAdapter_UnknownErrorsFromCallAreForwarded(t *testing.T) {
```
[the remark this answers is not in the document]
simonlechner: This could be any instance of an error, therefore arbitrary many different errors are handled that way. The test just gives an example.

go/geth_adapter/adapter_test.go
```diff
@@ -548,6 +538,180 @@ func TestRunContextAdapter_Run(t *testing.T) {
...
+		ChainID:       big.NewInt(42),
+		IstanbulBlock: big.NewInt(42),
+	}
+	blockContext := geth.BlockContext{
+		BlockNumber: big.NewInt(24),
+	}
+	evm := geth.NewEVM(blockContext, nil, chainConfig, geth.Config{})
+	adapter := &runContextAdapter{evm: evm}
+	_, err := adapter.Call(tosca.Call, tosca.CallParameters{})
+	require.Error(t, err, "unsupported revision")
```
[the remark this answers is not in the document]
facuMH: I agree. My comment was only a suggestion for a follow-up.

facuMH (approved):

HerbertJordan (approved):
Looks good!

PR #88, opened 2025-06-05
Fix same name receiver

facuMH:
This PR fixes findings by `staticcheck` rule ST1016.
This rule enforces that all methods with the same receiver type, use the same name to reference it.


simonlechner (approved):
Thank you!

PR #89, opened 2025-06-05
Fix documentation of exported methods

facuMH:
This PR fixes findings by `staticcheck` rule ST1020.
This rule enforces that the documentation, if any, of an exported method must begin with the name of the method.


simonlechner (approved):
Nice!

PR #90, opened 2025-06-05
Fix documentation of exported structs

facuMH:
This PR fixes findings by `staticcheck` rule ST1021.
This rule enforces that the documentation, if any, of an exported struct must begin with the name of the struct.


simonlechner (approved):
Thank you for fixing these!

PR #93, opened 2025-06-06
Make if-else into switch

facuMH:
This PR fixes findings by staticcheck rule QF1003.
This rule reports if/else-if blocks that could be turned into switches. The conversion is simple and direct.

 - [x] Run CT call rules


simonlechner (approved):
Looks good, I like that the `staticGas` is now set explicit for revisions newer than Berlin.

PR #94, opened 2025-06-06
Replace `strings.Replace` with `strings.ReplaceAll`

facuMH:
This PR fixes findings by `staticcheck` rule `QF1004`.
This rule reports uses of `strings.Replace` when the last argument is `-1` means it should replace all occurrences. Instead the method `strings.ReplaceAll` should be call, as it show in clearer manner the intent.

This rule will be activated with #86


simonlechner (approved):
Nice improvement in readability.

PR #96, opened 2025-06-06
Merge conditional assignment into variable declaration

facuMH:
This PR fixes findings by `staticcheck` rule `QF1007`.
This rule reports declarations of variables followed by an if that changes the variable value. 


simonlechner (commented):
It is very confusing to see a matching name and function name but in the assignment the value is negated. Please add a comment that it only returns true if it is the first call to selfdestruct.

PR #97, opened 2025-06-06
Remove unnecessary embedded fields

facuMH:
This PR fixes findings by `staticcheck` rule `QF1008`.
In `Go` when a `struct` has an *embedded* field, it allows the outer struct to call on methods and fields of the inner one, without actually naming the embedded struct.
This rule reports those cases where calling the embedded field can be omitted to simply call its methods.


simonlechner (approved):
Do be honest I am not a big fan of this rule, since it is no longer directly clear where the member field/function go introduced. But I think we should stick to idiomatic go so this is good to be merged.

PR #99, opened 2025-06-10
Remove 'deprecated' from `tosca.TransactionContext` methods

facuMH:
This PR fixes findings by `staticcheck` rule `QF1019`.
This rule reports usage of deprecated functions. 
The deprecated functions of `tosca.TransactionContext` are not meant to be deleted, since they are needed to maintain the geth reference implementation, but future implementations should not use (nor need) these "deprecated" functions. 
A proposal that should be investigated in the future is to make a `tosca.ExtendedTransactionContext`, that would inherit the necessary methods from `tosca.TransactionContext` and add the "deprecated" ones. 


simonlechner (approved):
Looks good for now. Refactoring the block processing interfaces is definitely a TODO.

PR #105, opened 2025-06-24
Geth adapter: remove debug feature

simonlechner:
The debug feature inside the geth adapter is rarely used and only prints fields which can easily be checked out in a debugger, therefore it is removed.


HerbertJordan (approved):
Thanks!

PR #109, opened 2025-06-26
Geth adapter: remove panics and add unit tests

simonlechner:
This PR removes panics and adds unit tests for the max call depth check and `...Log` functions.


HerbertJordan (approved):
Thanks for extending the test coverage, removing the panic calls!

PR #113, opened 2025-07-30
Processor: Add prague support

simonlechner:
This PR adds Prague support to the processor types and geth reference implementation by adding `SetCodeAuthorizations` and refactoring the `MakeChainConfig` function.

### DIFF
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -238,9 +238,10 @@ func transactionToMessage(transaction tosca.Transaction, gasPrice tosca.Value, b
 				ChainID: *uint256.NewInt(0).SetBytes(authorization.ChainID[:]),
 				Address: common.Address(authorization.Address),
 				Nonce:   authorization.Nonce,
-				V:       authorization.V,
-				R:       *uint256.NewInt(0).SetBytes(authorization.R[:]),
-				S:       *uint256.NewInt(0).SetBytes(authorization.S[:]),
+
+				V: authorization.V,                                 //nolint:all
+				R: *uint256.NewInt(0).SetBytes(authorization.R[:]), //nolint:all
+				S: *uint256.NewInt(0).SetBytes(authorization.S[:]), //nolint:all
 			}
 		}
 	}
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -239,6 +239,7 @@ func transactionToMessage(transaction tosca.Transaction, gasPrice tosca.Value, b
 				Address: common.Address(authorization.Address),
 				Nonce:   authorization.Nonce,
 
+				// Ignore linting due to the deprecation of the V, R, S fields.
 				V: authorization.V,                                 //nolint:all
 				R: *uint256.NewInt(0).SetBytes(authorization.R[:]), //nolint:all
 				S: *uint256.NewInt(0).SetBytes(authorization.S[:]), //nolint:all
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -239,10 +239,9 @@ func transactionToMessage(transaction tosca.Transaction, gasPrice tosca.Value, b
 				Address: common.Address(authorization.Address),
 				Nonce:   authorization.Nonce,
 
-				// Ignore linting due to the deprecation of the V, R, S fields.
-				V: authorization.V,                                 //nolint:all
-				R: *uint256.NewInt(0).SetBytes(authorization.R[:]), //nolint:all
-				S: *uint256.NewInt(0).SetBytes(authorization.S[:]), //nolint:all
+				V: authorization.V,
+				R: *uint256.NewInt(0).SetBytes(authorization.R[:]),
+				S: *uint256.NewInt(0).SetBytes(authorization.S[:]),
 			}
 		}
 	}
--- a/go/processor/geth/processor.go
+++ b/go/processor/geth/processor.go
@@ -230,18 +230,34 @@ func transactionToMessage(transaction tosca.Transaction, gasPrice tosca.Value, b
 		})
 	}
 
+	var authorizations []types.SetCodeAuthorization
+	if transaction.AuthorizationList != nil {
+		authorizations = make([]types.SetCodeAuthorization, len(transaction.AuthorizationList))
+		for idx, authorization := range transaction.AuthorizationList {
+			authorizations[idx] = types.SetCodeAuthorization{
+				ChainID: *uint256.NewInt(0).SetBytes(authorization.ChainID[:]),
+				Address: common.Address(authorization.Address),
+				Nonce:   authorization.Nonce,
+				V:       authorization.V,
+				R:       *uint256.NewInt(0).SetBytes(authorization.R[:]),
+				S:       *uint256.NewInt(0).SetBytes(authorization.S[:]),
+			}
+		}
+	}
+
 	return &core.Message{
-		From:          common.Address(transaction.Sender),
-		To:            (*common.Address)(transaction.Recipient),
-		Nonce:         transaction.Nonce,
-		Value:         transaction.Value.ToBig(),
-		GasLimit:      uint64(transaction.GasLimit),
-		GasPrice:      gasPrice.ToBig(),
-		GasFeeCap:     transaction.GasFeeCap.ToBig(),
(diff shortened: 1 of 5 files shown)


go/tosca/processor.go
```diff
@@ -24,18 +24,27 @@ type Processor interface {
...
+	AuthorizationList []SetCodeAuthorization
+}
 
+type SetCodeAuthorization struct {
+	ChainID Word
+	Address Address
+	Nonce   uint64
+	V       uint8
+	R       Word
+	S       Word
```
HerbertJordan, line 47: If you pass those fields, being the components of a signature, the processor is required to extract the signer from those. This is not a cheap operation.

I would suggest to pass the `Sender` as an `Address` explicitly here, analog to the `Transaction` above, where we are also not passing along the transaction signature but the decoded sender. This allows the reuse of the signature.

If you need those fields to be forwarded to the `geth` processor, they should be marked deprecated.

Wdyt?

go/tosca/processor.go
```diff
@@ -24,18 +24,27 @@ type Processor interface {
...
+	GasLimit          Gas           // the maximum amount of gas that can be used by the transaction
+	GasFeeCap         Value         // the amount of network currency the sender is willing to pay for one gas unit
+	GasTipCap         Value         // the amount of network currency the sender is willing to pay for one unit as a priority fee
+	BlobGasFeeCap     Value         // the amount of network currency the sender is willing to pay for blob gas
+	BlobHashes        []Hash        // the hashes of the blobs for this transaction
+	AccessList        []AccessTuple // the list of accounts and storage slots expected to be accessed
+	AuthorizationList []SetCodeAuthorization
+}
 
+type SetCodeAuthorization struct {
```
HerbertJordan, line 41: nit: document

HerbertJordan (commented):
The adaptation of the geth processor looks good, but the type used for Tosca's processor interface should be reconsidered to not require the processor the decryped the signature.

facuMH (commented):
This PR looks good but I have a couple of comments

facuMH (dismissed):

go/tosca/processor.go
HerbertJordan: It might actually be even beyond the scope of the transaction processor to verify these signatures. At some point we need to clarify those responsibilities.

go/tosca/processor.go
```diff
@@ -24,18 +24,33 @@ type Processor interface {
...
+	BlobHashes        []Hash        // the hashes of the blobs for this transaction
+	AccessList        []AccessTuple // the list of accounts and storage slots expected to be accessed
+	AuthorizationList []SetCodeAuthorization
+}
+
+// SetCodeAuthorization contains the information required for EIP-7702 set code transactions to
+// enable externally owned accounts (EOAs) to reference code and execute within their own context.
+type SetCodeAuthorization struct {
+	ChainID Word    // the chain ID for which this authorization is valid
+	Address Address // the target address of the delegation
```
HerbertJordan, line 45: There is a design choice to be made here: since a signature might be invalid, making it impossible to recover an Address from it, there can be authorizations for which no reasonable address can be extracted. However, we need to be able to forward them to the processor, since gas needs to be charged for them.

We have at least the following options:
- Option A: we use the signature V, R, and S, as geth does it; this, however, requires signature verification inside the processor, and since we pass the fields explicitly, we either have the inherent signature decoding computation costs in the processor (which might not be that heavy, depending on how extensive authorizations are used), or need to depend on some side-channel signature caching to eliminate the overhead
- Option B: make this field an optional by typing it as a `*Address`; if it is `nil`, the signature extraction failed; we could force all code to deal with the fact that this is an optional, my making it an actual `Optional[Address]`, but for this we would have to import an extra library like Serse's [optional](https://github.com/0xsoniclabs/serse/tree/main/lib/optional).
- Option C: have a "valid" boolean, indicating whether this authorization should be attempted to be processed or skipped;

I would lean towards option B; it is the simplest, if the `optional.T[Address]` is used, code is forced to deal with it. Wdyt?

go/processor/geth/processor_test.go
```diff
@@ -94,3 +97,123 @@ func TestGethProcessor_ConfigAddsStateContract(t *testing.T) {
...
+		Recipient:     &tosca.Address{0x02},
+		Nonce:         42,
+		Value:         tosca.NewValue(1000),
+		GasLimit:      21000,
+		GasFeeCap:     tosca.NewValue(50),
+		GasTipCap:     tosca.NewValue(2),
+		Input:         []byte{0x42, 0x42},
+		BlobGasFeeCap: tosca.NewValue(0),
+	}
+	gasPrice := tosca.NewValue(50)
```
HerbertJordan, line 113: This price should be something different than the `GasFeeCap` to make sure the `GasPrice` in the result is taken from this parameter and not from the `GasFeeCap`.

go/tosca/processor.go
simonlechner: For our own implementation, we should evaluate whether the verification should be part of the processor, but the types are required for the geth reference implementation.

go/processor/geth/processor_test.go
```diff
@@ -94,3 +97,123 @@ func TestGethProcessor_ConfigAddsStateContract(t *testing.T) {
...
+	if auth.ChainID.Cmp(uint256.NewInt(1)) != 0 {
+		t.Errorf("ChainID mismatch: got %v", auth.ChainID)
+	}
+	if auth.Address != common.Address(tx.AuthorizationList[0].Address) {
+		t.Errorf("Address mismatch")
+	}
+	if auth.Nonce != tx.AuthorizationList[0].Nonce {
+		t.Errorf("Nonce mismatch")
+	}
+	if auth.V != tx.AuthorizationList[0].V { //nolint:all
```
HerbertJordan, line 210: nit: For R and S you are avoiding the deprecated linter warning by comparing against the actual value, not the value from the input. So maybe here (unless you remove the deprecated state as mentioned in another comment):

```suggestion
	if auth.V != 42 {
```

go/tosca/processor.go
```diff
@@ -24,18 +24,33 @@ type Processor interface {
...
+	Nonce   uint64  // the nonce of the signer, used to prevent replay
 
+	// Deprecated: signature V value required by the geth processor.
```
HerbertJordan, line 48: This might have been a bad suggestion of mine. Marking this deprecated will cause us lots of linter issues, and people will just add `nolint` comments, which isn't great.

The problem is that there is no full replacement of those as long as we want to support the geth processor fallback (unless we patch our geth-fork accordingly). So while it is not pretty, we have to keep those as regular fields (not deprectated), containing redundant information to the `Address` field.

go/processor/geth/processor.go
```diff
@@ -230,18 +230,36 @@ func transactionToMessage(transaction tosca.Transaction, gasPrice tosca.Value, b
...
+		for idx, authorization := range transaction.AuthorizationList {
+			authorizations[idx] = types.SetCodeAuthorization{
+				ChainID: *uint256.NewInt(0).SetBytes(authorization.ChainID[:]),
+				Address: common.Address(authorization.Address),
+				Nonce:   authorization.Nonce,
+
+				// Ignore linting due to the deprecation of the V, R, S fields.
+				V: authorization.V,                                 //nolint:all
+				R: *uint256.NewInt(0).SetBytes(authorization.R[:]), //nolint:all
+				S: *uint256.NewInt(0).SetBytes(authorization.S[:]), //nolint:all
```
HerbertJordan, line 245: => see comment on not making those fields deprecated and please re-enable the linter here

HerbertJordan (approved):

go/processor/geth/processor_test.go
```diff
@@ -110,7 +110,7 @@ func TestTransactionToMessage_BasicFields(t *testing.T) {
 		Input:         []byte{0x42, 0x42},
 		BlobGasFeeCap: tosca.NewValue(0),
 	}
-	gasPrice := tosca.NewValue(50)
+	gasPrice := tosca.NewValue(40)
```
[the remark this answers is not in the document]
facuMH: maybe add a comment `// lower than basefee` or something similar

facuMH (approved):
This PR adds authorizatioin list support to the processor. 
It also updates the geth make config to support prague.

PR #114, opened 2025-07-30
Floria: precompiled contracts

simonlechner:
This PR adds an overflow check to the gas calculation of precompiled contracts and makes sure they are also handled correctly in a delegate call by using the code address.

### DIFF
--- a/go/processor/floria/precompiled_test.go
+++ b/go/processor/floria/precompiled_test.go
@@ -35,7 +35,7 @@ func TestPrecompiled_RightNumberOfContractsDependingOnRevision(t *testing.T) {
 		count := 0
 		for i := byte(0x01); i < byte(0x42); i++ {
 			address := test_utils.NewAddress(i)
-			_, isPrecompiled := getPrecompiledContract(address, test.revision)
+			isPrecompiled := isPrecompiled(address, test.revision)
 			if isPrecompiled {
 				count++
 			}
@@ -102,48 +102,5 @@ func TestPrecompiled_GasCostOverflowIsDetectedAndHandled(t *testing.T) {
 	}
 	if result.Success {
 		t.Errorf("expected failure, got success")
-	}
-}
-
-func TestIsPrecompiled(t *testing.T) {
-	tests := []struct {
-		name     string
-		revision tosca.Revision
-		address  tosca.Address
-		expected bool
-	}{
-		{
-			name:     "precompiled address in Istanbul",
-			revision: tosca.R07_Istanbul,
-			address:  test_utils.NewAddress(0x01),
-			expected: true,
-		},
-		{
-			name:     "non-precompiled address in Istanbul",
-			revision: tosca.R07_Istanbul,
-			address:  test_utils.NewAddress(0x20),
-			expected: false,
-		},
-		{
-			name:     "precompiled address in Cancun",
-			revision: tosca.R13_Cancun,
-			address:  test_utils.NewAddress(0x0a),
-			expected: true,
-		},
-		{
-			name:     "non-precompiled address in Cancun",
-			revision: tosca.R13_Cancun,
-			address:  test_utils.NewAddress(0x30),
-			expected: false,
-		},
-	}
-
-	for _, tc := range tests {
-		t.Run(tc.name, func(t *testing.T) {
-			got := isPrecompiled(tc.address, tc.revision)
-			if got != tc.expected {
-				t.Errorf("isPrecompiled(%v, %v) = %v; want %v", tc.address, tc.revision, got, tc.expected)
-			}
-		})
 	}
 }
--- a/go/processor/floria/precompiled_test.go
+++ b/go/processor/floria/precompiled_test.go
@@ -80,17 +80,11 @@ func TestPrecompiled_AddressesAreHandledCorrectly(t *testing.T) {
 			}
 
 			result, err := runPrecompiledContract(test.revision, input, test.address, test.gas)
-			if test.success && err != nil {
-				t.Errorf("unexpected error, want nil, got %v", err)
-				if !result.Success {
-					t.Errorf("expected success, but got %v", result.Success)
-				}
-			}
-			if !test.success && err == nil {
-				t.Errorf("expected an error, but got nil")
-				if result.Success {
-					t.Errorf("expected failure, but got success %v", result.Success)
-				}
+			if test.success {
+				require.NoError(t, err)
+				require.True(t, result.Success)
+			} else {
+				require.Error(t, err)
 			}
 		})
 	}
--- a/go/processor/floria/precompiled_test.go
+++ b/go/processor/floria/precompiled_test.go
@@ -83,3 +83,67 @@ func TestPrecompiled_AddressesAreHandledCorrectly(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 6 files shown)


go/processor/floria/precompiled_test.go
```diff
@@ -83,3 +83,67 @@ func TestPrecompiled_AddressesAreHandledCorrectly(t *testing.T) {
...
+}
+
+func TestIsPrecompiled(t *testing.T) {
+	tests := []struct {
+		name     string
+		revision tosca.Revision
+		address  tosca.Address
+		expected bool
+	}{
+		{
```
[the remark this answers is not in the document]
simonlechner: There is already a [test](https://github.com/0xsoniclabs/tosca/blob/f020a558b11e5eb35c1ca99ebdc5847788f0c72d/go/processor/floria/precompiled_test.go#L21) which ensures that the right number of precompiled contracts are available in each revision. I adapted the test and removed the one added in this PR.

go/processor/floria/precompiled_test.go
```diff
@@ -83,3 +83,24 @@ func TestPrecompiled_AddressesAreHandledCorrectly(t *testing.T) {
 		})
 	}
 }
+
+func TestPrecompiled_GasCostOverflowIsDetectedAndHandled(t *testing.T) {
+	data := []byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
HerbertJordan, line 88: nit: this test needs a bit of explanation; also, it should make sure that the execution actually failed for the claimed reasons, namely the detection of a gas overflow.

go/processor/floria/run_context_test.go
```diff
@@ -382,3 +383,57 @@ func TestIncrementNonce(t *testing.T) {
...
+				CodeAddress: test.codeAddress,
+				Value:       tosca.NewValue(0),
+				Gas:         1000,
+				Input:       []byte{},
+			})
+
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+			if result.Success {
```
HerbertJordan, line 430: Reading the test, I am not sure why the execution is supposed to fail ...

go/processor/floria/precompiled_test.go
simonlechner: I added documentation to the input and refactored the precompile contracts runner to be able to check errors properly.

go/processor/floria/run_context_test.go
simonlechner: I added an input which makes the calls succeed to the test.

go/processor/floria/run_context.go
```diff
@@ -61,36 +61,31 @@ func (r runContext) executeCall(kind tosca.CallKind, parameters tosca.CallParame
...
-	if r.blockParameters.Revision >= tosca.R09_Berlin &&
-		!isPrecompiled(recipient, r.blockParameters.Revision) &&
-		!isStateContract(recipient) &&
-		!r.AccountExists(recipient) &&
-		parameters.Value.Cmp(tosca.Value{}) == 0 {
+	isStateContract := isStateContract(parameters.CodeAddress)
+	isPrecompiled := isPrecompiled(parameters.CodeAddress, r.blockParameters.Revision)
+
+	if kind == tosca.Call && r.blockParameters.Revision >= tosca.R09_Berlin && !isPrecompiled &&
+		!isStateContract && !r.AccountExists(parameters.Recipient) && parameters.Value.Cmp(tosca.Value{}) == 0 {
```
facuMH, line 68: nit:
```suggestion
	if kind == tosca.Call &&
	 r.blockParameters.Revision >= tosca.R09_Berlin &&
	 !isPrecompiled &&
 	 !isStateContract &&
 	 !r.AccountExists(parameters.Recipient) &&
 	 parameters.Value.Cmp(tosca.Value{}) == 0 {
```
this is a crowded condition. I think separating it like this could help reading it. wdyt?

facuMH (approved):
a few minor comments

go/processor/floria/precompiled_test.go
```diff
@@ -73,9 +74,14 @@ func TestPrecompiled_AddressesAreHandledCorrectly(t *testing.T) {
...
-			result, isPrecompiled := handlePrecompiledContract(test.revision, input, test.address, test.gas)
+			isPrecompiled := isPrecompiled(test.address, test.revision)
 			if isPrecompiled != test.isPrecompiled {
-				t.Errorf("unexpected precompiled, want %v, got %v", test.isPrecompiled, isPrecompiled)
+				t.Fatalf("unexpected precompiled, want %v, got %v", test.isPrecompiled, isPrecompiled)
+			}
+
+			result, err := runPrecompiledContract(test.revision, input, test.address, test.gas)
+			if test.success && err != nil {
+				t.Errorf("unexpected error, want nil, got %v", err)
```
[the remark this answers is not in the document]
simonlechner: I updated the checks, now stressing both the error and the result.

go/processor/floria/run_context.go
```diff
@@ -61,10 +61,14 @@ func (r runContext) executeCall(kind tosca.CallKind, parameters tosca.CallParame
...
+	isPrecompiled := isPrecompiled(parameters.CodeAddress, r.blockParameters.Revision)
+
+	if kind == tosca.Call &&
+		r.blockParameters.Revision >= tosca.R09_Berlin &&
+		!isPrecompiled &&
+		!isStateContract &&
+		!r.AccountExists(parameters.Recipient) &&
 		parameters.Value.Cmp(tosca.Value{}) == 0 {
 		return tosca.CallResult{Success: true, GasLeft: parameters.Gas}, nil
 	}
```
HerbertJordan, line 74: This looks like an optimization of a call into a non-existing account. Can this be removed without changing anything?

go/processor/floria/state_contract.go
```diff
@@ -83,24 +83,24 @@ func isStateContract(address tosca.Address) bool {
 	return address == StateContractAddress()
 }
 
-// handleStateContract is a reworked version of the original function from the Opera client.
+// runStateContract is a reworked version of the original function from the Opera client.
```
HerbertJordan, line 86: nit: please aim for self-contained documentation, not assuming knowledge of third-party code from the perspective of the Tosca project.

go/processor/floria/state_contract.go
```diff
@@ -83,24 +83,24 @@ func isStateContract(address tosca.Address) bool {
...
 	state tosca.WorldState,
 	sender tosca.Address,
 	receiver tosca.Address,
 	input []byte,
 	gas tosca.Gas,
-) (tosca.CallResult, bool) {
+) tosca.CallResult {
 	if receiver != StateContractAddress() {
-		return tosca.CallResult{}, false
+		return tosca.CallResult{}
```
HerbertJordan, line 97: Curiously, this used to return "false" if the recipient/receiver is of the wrong address. In such a case, in the old code, the state-contract invocation is ignored, since `isStatePrecompiled`, the second return value, is false, and the regular call processing is performed.

In the new version given here, the recipient is ignored, and the call fails. This looks like a change in semantic.

PR #115, opened 2025-07-30
Floria: Add Blob support

simonlechner:
Floria was originally developed with the fantom substate, this PR adds blobs which were introduced in the Cancun revision.


HerbertJordan (commented):
I am having a difficult time lining up code changes with tests in this PR.

What is your development and evaluation method for these? Is this code a development version / prototype or is the aim to get this to production quality as is?

go/processor/floria/processor.go
```diff
@@ -302,3 +331,31 @@ func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gas
...
 }
+
+func blobCheck(transaction tosca.Transaction, blockParameters tosca.BlockParameters) error {
+	if transaction.BlobHashes != nil {
+		if transaction.Recipient == nil {
+			return fmt.Errorf("blob transactions need to have a existing recipient")
+		}
+		if len(transaction.BlobHashes) == 0 {
+			return fmt.Errorf("missing blob hashes")
+		}
```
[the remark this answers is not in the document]
simonlechner: [Ethereum](https://github.com/0xsoniclabs/go-ethereum/blob/bbe492095a30fe79d6bd81a360ee73680308c7d7/core/state_transition.go#L372) does not support blob transactions without any blobhashes. Sonic is using the geth processor, it also won't accept empty blob transactions.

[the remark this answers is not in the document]
HerbertJordan: I am running a scan through the block history whether there are blob transactions on the Sonic network.

go/processor/floria/processor.go
```diff
@@ -288,9 +294,14 @@ func calculateSetupGas(transaction tosca.Transaction) tosca.Gas {
 	return tosca.Gas(gas)
 }
 
-func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gasPrice tosca.Value) error {
+func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gasPrice tosca.Value, blobBaseFee tosca.Value) error {
```
HerbertJordan, line 297: Should `blobBaseFee` be renamed to `blobGasPrice`?

go/processor/floria/processor.go
```diff
@@ -302,3 +313,31 @@ func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gas
...
+			// Perform kzg4844 valid version check
+			if len(hash) != 32 || hash[0] != 0x01 {
+				return fmt.Errorf("blob with invalid hash version")
+			}
+		}
+
+	}
+
+	if blockParameters.Revision >= tosca.R13_Cancun && len(transaction.BlobHashes) > 0 {
+		if transaction.BlobGasFeeCap == (tosca.Value{}) {
```
HerbertJordan, line 335: Is the zero value the "no-cap" by definition of the specification or an artifact of the implementation?

go/processor/floria/processor_test.go
```diff
@@ -169,26 +169,31 @@ func TestProcessor_EoaCheck(t *testing.T) {
 }
 
 func TestProcessor_BuyGas(t *testing.T) {
-	balance := uint64(1000)
+	balance := uint64(1000000)
 	gasLimit := uint64(100)
 	gasPrice := uint64(2)
+	blobBaseFee := uint64(1)
```
HerbertJordan, line 175: nit: `blobGasPrice`?

go/processor/floria/processor_test.go
```diff
@@ -169,26 +169,31 @@ func TestProcessor_EoaCheck(t *testing.T) {
...
+	context.EXPECT().SetBalance(transaction.Sender, tosca.NewValue(newValue))
+	context.EXPECT().GetBalance(transaction.Sender).Return(tosca.NewValue(newValue))
 
-	err := buyGas(transaction, context, tosca.NewValue(gasPrice))
+	err := buyGas(transaction, context, tosca.NewValue(gasPrice), tosca.NewValue(blobBaseFee))
 	if err != nil {
 		t.Errorf("buyGas returned an error: %v", err)
 	}
-	if context.GetBalance(transaction.Sender).Cmp(tosca.NewValue(balance-gasLimit*gasPrice)) != 0 {
+	if context.GetBalance(transaction.Sender).Cmp(tosca.NewValue(newValue)) != 0 {
```
HerbertJordan, line 196: If I read this correctly, this line is only testing whether the expectation of line 190 is set up correctly. If so, it should be removed.

go/processor/floria/processor_test.go
```diff
@@ -504,3 +509,102 @@ func TestProcessor_AccessListIsNotCreatedIfTransactionHasNone(t *testing.T) {
 
 	setUpAccessList(transaction, context, tosca.R09_Berlin)
 }
+
+func TestProcessor_blobCheckReturnsErrors(t *testing.T) {
+	tests := map[string]struct {
+		transaction tosca.Transaction
+		blockParams tosca.BlockParameters
+		expectError bool
```
HerbertJordan, line 517: nit: to make sure that tests fail for the expected reasons, you should check that the error message contains some expected issue report

HerbertJordan (approved):

go/processor/floria/processor_test.go
```diff
@@ -169,26 +169,31 @@ func TestProcessor_EoaCheck(t *testing.T) {
...
 
 	transaction := tosca.Transaction{
-		Sender:   tosca.Address{1},
-		GasLimit: tosca.Gas(gasLimit),
+		Sender:     tosca.Address{1},
+		GasLimit:   tosca.Gas(gasLimit),
+		BlobHashes: []tosca.Hash{{0x01}},
 	}
 
+	newValue := balance - (gasLimit*gasPrice +
```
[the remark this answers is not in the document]
simonlechner: I changed it to `newBalance`.

go/processor/floria/processor_test.go
```diff
@@ -504,3 +509,102 @@ func TestProcessor_AccessListIsNotCreatedIfTransactionHasNone(t *testing.T) {
...
+		blockParams tosca.BlockParameters
+		expectError bool
+	}{
+		"valid blob transaction pre cancun": {
+			transaction: tosca.Transaction{
+				Recipient:  &tosca.Address{1},
+				BlobHashes: []tosca.Hash{{1}},
+			},
+			blockParams: tosca.BlockParameters{Revision: tosca.R12_Shanghai},
+			expectError: false,
```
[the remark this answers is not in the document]
simonlechner: It is not possible to have a transaction with blobs before Cancun, the check is still going to be performed for older transactions. The test case is to ensure that they do not fail on any of these checks. I adapted the test.

go/processor/floria/processor.go
HerbertJordan: In geth this check is only skipped if the `NoBaseFee` flag is set in the configuration. Otherwise, it looks to me that 0 is a valid entry that should lead to an error if it is underpriced. This is not the case here.

The `NoBaseFee` flag would be set if the EVM is used in an RPC call with releaxed rules.

HerbertJordan (changes requested):

go/processor/floria/processor.go
simonlechner: I added the `NoBaseFee` flag with a TODO to wrap it in a struct. I want to add the config in a separate PR. I am also working on the buyGas function and will add the flag there as well.

go/processor/floria/processor.go
```diff
@@ -302,3 +314,31 @@ func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gas
...
+			if len(hash) != 32 || hash[0] != 0x01 {
+				return fmt.Errorf("blob with invalid hash version")
+			}
+		}
+
+	}
+
+	if blockParameters.Revision >= tosca.R13_Cancun && len(transaction.BlobHashes) > 0 {
+		if noBaseFee && transaction.BlobGasFeeCap == (tosca.Value{}) {
+			return nil // skip checks if no blob gas fee cap is set
```
HerbertJordan, line 337:
```suggestion
			return nil // skip checks if base fees should not be checked
```

go/processor/floria/processor.go
```diff
@@ -302,3 +314,31 @@ func buyGas(transaction tosca.Transaction, context tosca.TransactionContext, gas
 
 	return nil
 }
+
+func checkBlobs(transaction tosca.Transaction, blockParameters tosca.BlockParameters, noBaseFee bool) error {
```
HerbertJordan, line 318: Consider finding a better fitting name for `noBaseFee`. As it is now, it should be called `allwaysAcceptZeroBlobGasFeeCap` ...

HerbertJordan (approved):
As discussed offline, I consider this now a prototype implementation. Thus, I did not go in detail through all the test cases.

facuMH (dismissed):
This PR adds mechanism to account/charge for blobs during the processor run and testing. 
The main addition is the `checkBlobs` function and its testing.

go/processor/floria/processor.go
simonlechner: I removed the flag.

PR #117, opened 2025-07-31
WorldState: Add createContract method

simonlechner:
This PR adds a function `createContract` to the world state, originally it was expected that all cases can be handled with a single `createAccount` function, but this is not the case. There cases related to selfdestruct where the distinction during the creation is imported.


go/geth_adapter/adapter.go
```diff
@@ -382,9 +382,10 @@ func gethToVMErrors(err error, gas tosca.Gas) (tosca.CallResult, error) {
 }
 
 func (a *runContextAdapter) CreateAccount(addr tosca.Address) {
-	if !a.evm.StateDB.Exist(common.Address(addr)) {
```
facuMH, line 385: why is it no longer necessary to check whether the account exists or not ?

go/integration_test/interpreter/test_evm_test.go
```diff
@@ -0,0 +1,41 @@
...
+package interpreter_test
+
+import (
+	"testing"
+
+	"github.com/0xsoniclabs/tosca/go/tosca"
+	"go.uber.org/mock/gomock"
+)
+
+func TestScenario_CreateAccountAndContractCreateAnEmptySlot(t *testing.T) {
```
facuMH, line 20: I'm not sure if we should add tests for empty function. Or maybe it should be in a comment that this test is only checking that the function call should not change or trigger anything.

facuMH (commented):
This PR adds the `CreateContract` function to the `WorldState` interface and its implementations.

go/tosca/world_state.go
```diff
@@ -21,6 +21,7 @@ type WorldState interface {
 	AccountExists(Address) bool
 
 	CreateAccount(Address)
+	CreateContract(Address)
```
HerbertJordan, line 24: The aim of this interface is to provide a clean separation between the DB and the VM. Part of this is to have a clear semantic for all these operations. Things like `GetBalance` and `SetBalance` are quite self-explaining, `AccountExists` is borderline, but the difference between these functions (and their respective effects) needs to be clarified in this interface definition.

go/integration_test/interpreter/test_evm_test.go
simonlechner: As discussed offline, there is currently no better way to test these functions in the context of the `test_evm`.

go/tosca/world_state.go
simonlechner: I added comments to all three functions.

HerbertJordan (changes requested):
Part of building our own processor is to not only figure out what it is doing, but why it is doing. If we just copy operation from a reference version, like in this case "CreateContract", without understanding what their purpose is, we will not understand the result and we will not be able to develop it freely.

Based on the added comments, I do not know what these functions are for, what the effect of their invocation is, when they should be called, and when not. It might also be that these functions are just artifacts of the Geth implementation, that can be handled differently. 

If this is supposed to be a product quality code, these questions need to be answered.

simonlechner:
It will take further investigation to see whether the `CreateContract` function is required.

PR #118, opened 2025-08-01
Remove strict project coverage threshold

wsodsong:
This PR removes strict project coverage target. Project coverage meansures coverage level of the entire repository. Previously, project coverage has not been activated. With the recent codecov plan change, project coverage takes effects. This condition may unnecessary prevent PR from merging.


simonlechner (approved):
I think it currently does not make sense to enforce a project wide coverage limit, we have some untested code, especially in the domain of testing and work in progress.

HerbertJordan (approved):

PR #119, opened 2025-08-01
Floria: Cancun support

simonlechner:
This PR adds the missing features and fixes to support revisions up to Cancun.
- Snapshot rollback in case of errors after gas has been bought
- Ensure the gasFeeCap is smaller than the gasTipCap
- Warm coinbase (introduced in Shanghai)
- Gas calculation for init code (adapted in Shainghai)

### DIFF
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -23,6 +23,7 @@ const (
 	TxDataZeroGasEIP2028      = 4
 	TxAccessListAddressGas    = 2400
 	TxAccessListStorageKeyGas = 1900
+	InitCodeWordGas           = 2
 
 	createGasCostPerByte = 200
 	maxCodeSize          = 24576
@@ -56,11 +57,11 @@ func (p *processor) Run(
 		Success: false,
 		GasUsed: transaction.GasLimit,
 	}
+	snapshot := context.CreateSnapshot()
 	gasPrice, err := calculateGasPrice(blockParameters.BaseFee, transaction.GasFeeCap, transaction.GasTipCap)
 	if err != nil {
 		return errorReceipt, err
 	}
-	gas := transaction.GasLimit
 
 	if nonceCheck(transaction.Nonce, context.GetNonce(transaction.Sender)) != nil {
 		return tosca.Receipt{}, nil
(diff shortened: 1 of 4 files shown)


go/processor/floria/processor.go
```diff
@@ -99,7 +104,7 @@ func (p *processor) Run(
 	}
 
 	if blockParameters.Revision >= tosca.R09_Berlin {
-		setUpAccessList(transaction, &runContext, blockParameters.Revision)
+		setUpAccessList(transaction, &runContext, blockParameters.Revision, blockParameters.Coinbase)
 	}
 
 	callParameters := callParameters(transaction, gas)
```
facuMH, line 110: if the call in line 117 fails, this is considered a fail transaction and there is no need for restore transaction, correct ?

go/processor/floria/processor_test.go
```diff
@@ -39,6 +39,40 @@ func TestProcessorRegistry_InitProcessor(t *testing.T) {
 	}
 }
 
+func TestProcessor_Run_PassingExecution(t *testing.T) {
```
facuMH, line 42:
```suggestion
func TestProcessor_Run_SuccessfulExecution(t *testing.T) {
```

facuMH (commented):
This PR implements changes introduced for the Cancun fork and tests for the corresponding new behaviours. 
In principle looks good, but I have a couple of comments before approving.

go/processor/floria/processor_test.go
```diff
@@ -477,16 +534,17 @@ func TestProcessor_SetUpAccessList(t *testing.T) {
...
 	}
 	context.EXPECT().AccessAccount(sender)
 	context.EXPECT().AccessAccount(recipient)
 	context.EXPECT().AccessAccount(accessListAddress)
 	context.EXPECT().AccessStorage(accessListAddress, tosca.Key{1})
 	context.EXPECT().AccessStorage(accessListAddress, tosca.Key{2})
+	context.EXPECT().AccessAccount(coinbase)
 
-	setUpAccessList(transaction, context, tosca.R09_Berlin)
+	setUpAccessList(transaction, context, tosca.R13_Cancun, coinbase)
```
HerbertJordan, line 547: It is not enough to use one revision to test the code. This is basically just adding code coverage, but it does not check that other revisions are handled correctly.

HerbertJordan (changes requested):
This PR covers multiple independent changes:
- extra init code costs in Shanghai and beyond
- the creation of a snapshot before buying gas and the rollback if buying gas failed
- the rollback if there is not enough gas for the transaction setup
- the implicit transaction consistency check in `calculateGasPrice` making sure that the tip cap is not bigger than the fee cap
- the inclusion of the coin base as a warm address during transaction setup
- the inclusion of init-code costs in the setup costs after shanghai
- the skipping of value transfers for `CallCode` calls
- the abort of CREATE calls if the created account is not empty
- the implicit creation of accounts after CREATE calls if they do not exit

All of these are substantial changes that need proper reviews and testing. Given the state of the present tests, it does not seem to me that all of these features are covered sufficiently with tests.

go/processor/floria/processor.go
simonlechner: The snapshot only needs to be restored if the gas has been bought, but the transaction has not been executed due to an error in any of the pre checks.

go/processor/floria/run_context.go
```diff
@@ -69,7 +69,7 @@ func (r runContext) executeCall(kind tosca.CallKind, parameters tosca.CallParame
 		return tosca.CallResult{Success: true, GasLeft: parameters.Gas}, nil
 	}
 
-	if kind == tosca.Call || kind == tosca.CallCode {
```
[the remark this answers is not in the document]
simonlechner: The balance transfer inside a callcode would transfer balance from the sender to itself, this can be skipped.

HerbertJordan (approved):
As discussed offline, I am approving this in the assumption that this is a prototype and it will be reworked.

facuMH (approved):
Considering this is a prototype, is fine as is.

PR #121, opened 2025-08-12
Floria: Refactor gas and balance handling

simonlechner:
This PR improves the buying, calculation and handling of gas. A new check is introduced to ensure the sender has a sufficient amount of balance to perform the transaction. Due to this check, it is no longer required to check the balance while buying the gas.


facuMH (approved):
Thanks for the addition. This can be merged regardless of coverage since follow-up work will cover the missing lines.

PR #124, opened 2025-08-13
Floria: Refactor run function

simonlechner:
This PR splits floria's main `Run` function into four functions, making it much more readable and maintainable. In order to keep the review simple, this PR does not add any other changes other than the split into multiple functions.

### DIFF
--- a/go/processor/floria/processor.go
+++ b/go/processor/floria/processor.go
@@ -60,27 +60,31 @@ type Processor struct {
 	EthCompatible bool
 }
 
+// Run checks whether the transaction can be executed and applies it if possible.
+// It returns a receipt with the result of the transaction or an error in case the
+// transaction can not be executed.
 func (p *Processor) Run(
 	blockParameters tosca.BlockParameters,
 	transaction tosca.Transaction,
 	context tosca.TransactionContext,
 ) (tosca.Receipt, error) {
 
-	if err := preCheck(blockParameters, transaction, context); err != nil {
+	if err := checkTransaction(blockParameters, transaction, context); err != nil {
 		return tosca.Receipt{}, err
 	}
 
-	gasPrice, gas, err := buyGas(blockParameters, transaction, context, p.EthCompatible)
+	gasPrice, gas, err := calculateAvailableGas(blockParameters, transaction, context, p.EthCompatible)
 	if err != nil {
-		return tosca.Receipt{GasUsed: gas}, err
+		return tosca.Receipt{}, err
 	}
 
 	result, err := p.runTransaction(blockParameters, transaction, context, p.EthCompatible, gasPrice, gas)
 	if err != nil {
+		// An error here is due to an implementation bug or runtime error, the state has already been modified.
 		return tosca.Receipt{GasUsed: transaction.GasLimit}, err
 	}
 
-	gasUsed := refundGas(blockParameters, transaction, context, gasPrice, result, p.EthCompatible)
+	gasUsed := returnExcessGas(blockParameters, transaction, context, gasPrice, result, p.EthCompatible)
 
 	receipt := tosca.Receipt{
 		Success: result.Success,
@@ -96,7 +100,10 @@ func (p *Processor) Run(
 	return receipt, nil
 }
 
-func preCheck(
+// checkTransaction performs basic checks to ensure the validity of the transaction.
+// It ensures the nonce is correct, the transaction is from an EOA, blobs are valid,
+// and the init code size is not too large.
+func checkTransaction(
 	blockParameters tosca.BlockParameters,
 	transaction tosca.Transaction,
 	context tosca.TransactionContext,
@@ -120,7 +127,10 @@ func preCheck(
 	return nil
 }
 
-func buyGas(
+// calculateAvailableGas calculates the gasPrice and available gas for the transaction.
+// An error is returned if the gasPrice can not be calculated, the sender does not have
+// enough balance or the gasLimit is too low for the set up gas.
+func calculateAvailableGas(
 	blockParameters tosca.BlockParameters,
 	transaction tosca.Transaction,
 	context tosca.TransactionContext,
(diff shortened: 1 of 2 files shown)


go/processor/floria/processor.go
```diff
@@ -313,7 +363,7 @@ func calculateGasLeft(transaction tosca.Transaction, result tosca.CallResult, re
 	return gasLeft
 }
 
-func refundGas(context tosca.TransactionContext, sender tosca.Address, gasPrice tosca.Value, gasLeft tosca.Gas) {
+func refundGasInternal(context tosca.TransactionContext, sender tosca.Address, gasPrice tosca.Value, gasLeft tosca.Gas) {
```
LuisPH3, line 366: please add equivalent docstring as in `buyGasInternal`

go/processor/floria/processor.go
```diff
@@ -65,39 +65,95 @@ func (p *Processor) Run(
 	transaction tosca.Transaction,
 	context tosca.TransactionContext,
 ) (tosca.Receipt, error) {
```
LuisPH3, line 67: I think that this function is not very standard:

Ideally, if error is not nil, any other value has to be ignored.
In this function, both a valid receipt and an error are returned. The unexpected thing is that callers are expected to honor the returned receipt.

 Can we document this behavior?
 - what does it mean to return gas when buyGas returns an error? 
 - what does it mean to return gas limit when the tx returned an error? 
 
I am guessing that here the biggest question is:  what does it mean to return an error in this call ? 
if error means that the transaction failed: Then we may just want to return the receipt directly from this function. 
if an error means an internal error in the system, this means some serious problem in the interpreter and we can well just crash. 

wdyt?

go/processor/floria/processor.go
```diff
@@ -65,39 +65,95 @@ func (p *Processor) Run(
...
-	gasPrice, err := calculateGasPrice(blockParameters.BaseFee, transaction.GasFeeCap, transaction.GasTipCap)
+
+	if err := preCheck(blockParameters, transaction, context); err != nil {
+		return tosca.Receipt{}, err
+	}
+
+	gasPrice, gas, err := buyGas(blockParameters, transaction, context, p.EthCompatible)
+	if err != nil {
+		return tosca.Receipt{GasUsed: gas}, err
```
LuisPH3, line 75: this is odd. should not rely on gas being initialized.

I am guessing that `setupGas := calculateSetupGas(transaction, blockParameters.Revision)` does not belong inside buyGas. (notice that this is a static property of the transaction, intrinsic gas. It could be checked in preCheck) 


I am confused about returning some gas consumption here. See geth code here:
```go
func (st *stateTransition) execute() (*ExecutionResult, error) {
	// First check this message satisfies all consensus rules before
	// applying the message. The rules include these clauses
	//
	// 1. the nonce of the message caller is correct
	// 2. caller has enough balance to cover transaction fee(gaslimit * gasprice)
	// 3. the amount of gas required is available in the block
	// 4. the purchased gas is enough to cover intrinsic usage
	// 5. there is no overflow when calculating intrinsic gas
	// 6. caller has enough balance to cover asset transfer for **topmost** call

	// Check clauses 1-3, buy gas if everything is correct
	if err := st.preCheck(); err != nil {
		return nil, err
	}

	var (
		msg              = st.msg
		rules            = st.evm.ChainConfig().Rules(st.evm.Context.BlockNumber, st.evm.Context.Random != nil, st.evm.Context.Time)
		contractCreation = msg.To == nil
		floorDataGas     uint64
	)

	// Check clauses 4-5, subtract intrinsic gas if everything is correct
	gas, err := IntrinsicGas(msg.Data, msg.AccessList, msg.SetCodeAuthorizations, contractCreation, rules.IsHomestead, rules.IsIstanbul, rules.IsShanghai)
	if err != nil {
		return nil, err
	}
```
There is no cost, the tx just fails, I pressure that gas cost will be automatically set to the maximum.

go/processor/floria/processor.go
```diff
@@ -65,39 +65,95 @@ func (p *Processor) Run(
...
+	gasPrice, err := calculateGasPrice(blockParameters.BaseFee, transaction.GasFeeCap, transaction.GasTipCap)
+	if err != nil {
+		return tosca.Value{}, 0, fmt.Errorf("failed to calculate gas price: %w", err)
+	}
+
+	if err = balanceCheck(gasPrice, transaction, context.GetBalance(transaction.Sender), ethCompatible); err != nil {
+		return tosca.Value{}, 0, fmt.Errorf("failed balance check: %w", err)
 	}
 
 	setupGas := calculateSetupGas(transaction, blockParameters.Revision)
```
LuisPH3, line 138: this check could be moved to the preChecks

go/processor/floria/processor.go
```diff
@@ -65,39 +65,95 @@ func (p *Processor) Run(
...
+	if transaction.Recipient == nil {
+		receipt.ContractAddress = &result.CreatedAddress
 	}
 
-	if err = eoaCheck(transaction.Sender, context.GetCodeHash(transaction.Sender)); err != nil {
-		return tosca.Receipt{}, fmt.Errorf("failed EOA check: %w", err)
+	return receipt, nil
+}
+
+func preCheck(
```
LuisPH3, line 99: I suggest renaming this to "static" checks. or "sanity" checks. 
just to avoid aliasing to geth

LuisPH3 (changes requested):
please consider my comments, It could be that some of them are out of scope for the partitioning of the function. we can address these in a separate PR.

LuisPH3:
> please consider my comments, It could be that some of them are out of scope for the partitioning of the function. we can address these in a separate PR.

After reviewing the other PR, I realized that some of these gas values are being cleaned up. Nevertheless, I thought about something:

the tuple (receipt, error)  is redundant, any value (x, error)  has to be converted into `receipt{ notOk, maxGasCost}`, 
I realized that Geth has a type with can be complemented with an error: 
```go
// ExecutionResult includes all output after executing given evm
// message no matter the execution itself is successful or not.
type ExecutionResult struct {
	UsedGas    uint64 // Total used gas, not including the refunded gas
	MaxUsedGas uint64 // Maximum gas consumed during execution, excluding gas refunds.
	Err        error  // Any error encountered during the execution(listed in core/vm/errors.go)
	ReturnData []byte // Returned data from evm(function result or data supplied with revert opcode)
}
```
maybe we are missing some type here, returning a receipt and error is strange.

go/processor/floria/processor.go
simonlechner: I renamed the function to `checkTransaction`.

PR #127, opened 2025-08-20
Floria: factor out intpreter call

simonlechner:
The difference between the handling of calls and creates is quite big, but they use the same parameter set-up and call to the interpreter, therefore it is factored out into its own function.

### DIFF
--- a/go/processor/floria/run_context_test.go
+++ b/go/processor/floria/run_context_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/0xsoniclabs/tosca/go/tosca"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/stretchr/testify/require"
 	"go.uber.org/mock/gomock"
 )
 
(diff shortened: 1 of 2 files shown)


go/processor/floria/run_context_test.go
```diff
@@ -478,6 +479,193 @@ func TestRunContext_AccountIsOnlyCreatedIfItIsEmptyAndDoesNotExist(t *testing.T)
 	}
 }
 
+func TestRunContext_runInterpreterSelectsCodeBasedOnType(t *testing.T) {
```
facuMH, line 482: what if make an `interptrerMock` and expect the specific parameters for `interpreter.Run` ?

LuisPH3 (dismissed):
This PR factors the common setup of parameters to do an interpreter call.

go/processor/floria/run_context_test.go
simonlechner: As we discussed, I removed the result check to keep the attention on the checks inside the interpreter mock.

PR #128, opened 2025-08-21
Floria: add tests for runContext fields

simonlechner:
The `runContext` saves the current call depth and whether it is in a static context. These values need to be handled by the processor and are not passed by the interpreter to the next call. The `runContext` is currently passed to the interpreter by value, therefore no reset of these values is required. This PR mainly adds unit tests to ensure eventual future changes do not break the depth and static values.


go/processor/floria/run_context_test.go
```diff
@@ -1147,3 +1147,254 @@ func allCallTypes() []tosca.CallKind {
...
+			nestedContext, ok := nestedParams.Context.(runContext)
+			require.True(t, ok, "Context should be of type runContext")
+			require.True(t, nestedContext.static, "Static flag should be set in nested static call")
+			require.True(t, nestedParams.Static, "Static flag should be set in nested static call parameters")
+			return tosca.Result{Success: true}, nil
+		})
+		// Normal call inside top level static call should still be in static context
+		_, err = params.Context.Call(tosca.Call, parameters)
+		require.Nil(t, err, "Call should not return an error")
+		require.True(t, innerContext.static, "Static flag should be set after nested calls")
```
LuisPH3, line 1209: nit:
```suggestion
		require.True(t, innerContext.static, "Parent call context remains static after first nested call")
```

go/processor/floria/run_context_test.go
```diff
@@ -1147,3 +1147,254 @@ func allCallTypes() []tosca.CallKind {
...
+	// 0 -------------> 1 -------------> 2
+	// 0                1 <------------- 2
+	// 0                1                2
+	// 0                1      call      2
+	// 0                1 -------------> 2
+	// 0                1 <------------- 2
+	// 0 <------------- 1                2
+
+	interpreter.EXPECT().Run(gomock.Any()).DoAndReturn(func(params tosca.Parameters) (tosca.Result, error) {
+		innerContext, ok := params.Context.(runContext)
```
LuisPH3, line 1182: nit:
inner may not be the best name. Because you have a nested call, inner(most) is the nested call.
I think parentContext may be better, or topLevelContext to match the mentioned top-level Call

LuisPH3 (dismissed):
This PR uses call by value to avoid cleaning up resources after recursion.
Testing includes nice documentation for complicated setups.

some notes:

go/processor/floria/run_context_test.go
simonlechner: I added the call depth to the variable name, so they align with the ascii model.

go/processor/floria/run_context_test.go
simonlechner: I updated the comments to keep them consistent in all tests.

LuisPH3 (approved):
:+1

PR #130, opened 2025-08-25
Floria: clean up error handling

simonlechner:
A call can fail due to several reasons, different failures need to be addressed accordingly. This PR simplifies the handling of unsuccessful calls and adds unit testing to cover different cases.

### DIFF
--- a/go/processor/floria/run_context.go
+++ b/go/processor/floria/run_context.go
@@ -57,6 +57,8 @@ func (r runContext) executeCall(kind tosca.CallKind, parameters tosca.CallParame
 
 	snapshot := r.CreateSnapshot()
 	defer func() {
+		// For all returns with an error or an unsuccessful result,
+		// the snapshot will be restored.
 		if err != nil || !callResult.Success {
 			r.RestoreSnapshot(snapshot)
 		}
@@ -138,6 +140,8 @@ func (r runContext) executeCreate(kind tosca.CallKind, parameters tosca.CallPara
 	// created address. The nonce increment of the sender is not impacted.
 	snapshot := r.CreateSnapshot()
 	defer func() {
+		// For all returns with an error or an unsuccessful result,
+		// the snapshot will be restored.
 		if err != nil || !callResult.Success {
 			r.RestoreSnapshot(snapshot)
 		}


go/processor/floria/run_context.go
```diff
@@ -154,14 +153,18 @@ func (r runContext) executeCreate(kind tosca.CallKind, parameters tosca.CallPara
...
-		return tosca.CallResult{Output: result.Output, GasLeft: result.GasLeft, CreatedAddress: createdAddress}, nil
+		return tosca.CallResult{}, err
+	}
+	if !result.Success {
+		r.RestoreSnapshot(snapshot)
+		return tosca.CallResult{
+			Output:         result.Output,
+			GasLeft:        result.GasLeft,
+			GasRefund:      result.GasRefund,
+			CreatedAddress: createdAddress,
```
LuisPH3, line 166: why is created address not part of the value returned from `runInterpreter`?   
having to patch the return seems a little odd. 

If you manage to return the right value from the beginning, there is no need to create a new result instance here.

LuisPH3 (approved):

go/processor/floria/run_context.go
simonlechner: The interpreter should not know about and deal with the account creation, it does not make a difference whether the interpreter gets called in a create or a normal call. The interpreter's result does not contain a `createdAddress` field.

go/processor/floria/run_context.go
```diff
@@ -154,14 +153,18 @@ func (r runContext) executeCreate(kind tosca.CallKind, parameters tosca.CallPara
...
-
-		if !isRevert(result, err) {
-			// if the unsuccessful create was due to a revert, the result is still returned
-			return tosca.CallResult{}, err
-		}
-		return tosca.CallResult{Output: result.Output, GasLeft: result.GasLeft, CreatedAddress: createdAddress}, nil
+		return tosca.CallResult{}, err
+	}
+	if !result.Success {
+		r.RestoreSnapshot(snapshot)
```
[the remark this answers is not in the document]
LuisPH3: I agree about the one location, but I am concerned about maintainability. 
All the code we have seen so far deals with multiple exceptional cases and finicky patterns to handle all of them in if-else clauses. This builds up pretty fast to be difficult to understand and requires constant attention to corner cases. I would like to see patterns that attempt to guard against higher level conditions. For example:
instead of thinking:
- "when an account has value x and the scenario is Y then something has to happen"
let's think something like:
- "when the operation succeeds, a and b need to happen, if it does not, Y"  

The later one reads as:  a call is a transaction, if the call succeeds the changes get committed to DB. if not, rolled back. 

It would be really nice to see this reflected in the code. The current code is unfortunately a collection of corner cases

go/processor/floria/run_context.go
```diff
@@ -54,7 +54,14 @@ func (r runContext) executeCall(kind tosca.CallKind, parameters tosca.CallParame
 			return errResult, nil
 		}
 	}
+
 	snapshot := r.CreateSnapshot()
+	defer func() {
+		if err != nil || !callResult.Success {
+			r.RestoreSnapshot(snapshot)
+		}
+	}()
```
LuisPH3, line 63: nit: little text accompanying this code may help:


```go
    // A call to the VM is a transactional operation, if the call fails either by error or unsuccesulf execution:
    // changes to the stateDb are reverted
	snapshot := r.CreateSnapshot()
	defer func() {
		if err != nil || !callResult.Success {
			r.RestoreSnapshot(snapshot)
		}
	}()
```

LuisPH3 (dismissed):

PR #132, opened 2025-08-29
Floria: add config to processor

simonlechner:
This PR adds further configuration options to Floria. A new flag, `offChainExecution` is added which skips several checks in the case of a `eth_call` or `eth_simulate` simulation run. 
The state contract can now also be passed in the config, this allows simple changes to it.


go/processor/floria/processor.go
```diff
@@ -49,15 +49,43 @@ func init() {
...
+	Run(state tosca.WorldState, sender tosca.Address, receiver tosca.Address, input []byte, gas tosca.Gas) tosca.CallResult
+}
+
+// Config holds configuration options for the Floria processor.
+type Config struct {
+	// EthCompatible flag indicates whether the processor is compatible with Ethereum.
+	// The default is compatibility with Sonic/Fantom.
 	EthCompatible bool
+
+	// StateContracts allows to add special contracts that can not be implemented as
```
LuisPH3, line 78: "that can not" does not work very well.

```suggestion
	// StateContracts binds addresses to special contracts not deployed in the blockchain. 
```

go/processor/floria/processor.go
```diff
@@ -49,15 +49,43 @@ func init() {
...
 }
 
 // Processor implements the tosca.Processor interface for the Floria processor.
 type Processor struct {
-	Interpreter   tosca.Interpreter
+	Interpreter tosca.Interpreter
+	Config      Config
+}
+
+type StateContract interface {
```
LuisPH3, line 68: StateContract is an interface which allows injecting custom handling code for registered addresses. 
From the caller side, such addresses can be called as any other smart contract address.

go/processor/floria/processor.go
```diff
@@ -49,15 +49,43 @@ func init() {
...
+
+	// StateContracts allows to add special contracts that can not be implemented as
+	// classic smart contracts.
+	StateContracts map[tosca.Address]StateContract
+
+	// OffChainExecution flag indicates that the transaction does not have any effect
+	// on the state of the blockchain. It is used for simulation purposes.
+	// When enabled, EOA and nonce checks are skipped, and 0 is treated as a valid value for
+	// gasprice and blob gasprice.
+	OffChainExecution bool
```
LuisPH3, line 86: Maybe we should consider upgrading `Simulation` to be a concept in the processor. 
IMO: OffChain does not contribute to the meaning any more than simulation.
wdyt @HerbertJordan ?

go/processor/floria/processor.go
LuisPH3, line 68: nit: because this is generic over the actual execution, we can make the name a little more generic:
`BuiltInContract`

LuisPH3 (changes requested):
some naming comments

PR #134, opened 2025-09-01
Fix handling of snapshots of refund states in state DB adapter

HerbertJordan:
This PR fixes the creation and restoration of snapshots in the `geth_adaptor`'s StateDB as it is used by Tosca's geth reference processor implementation.

Before this fix, the implementation depended on the underlying system to produce distinct snapshot IDs for each call. However, this is not the case for some implementations, in particular the `Carmen` DB, which may reuse IDs.

With this fix, this problem got resolved.


simonlechner (approved):
Thanks for the fix.

PR #135, opened 2025-09-01
Support handling of internal transactions in geth processor

HerbertJordan:
This PR introduces a special-case handling for the gas price fee handling for internal transactions. In particular, internal transactions:
 - are charged the gas price as stated as their gas-fee-cap
 - are not bound by the block's base-fee limit; gas prices may be arbitrary low, including 0


simonlechner (approved):
Thank you for fixing this.

PR #136, opened 2025-09-09
CT: spec changes for completeness and determinism

simonlechner:
This PR adds changes found by the SMT completeness and determinism check.
- add `isCode(Pc))` condition to several rules
- remove over specification from several rules with failure effect
- fix `jump_not_taken` rule


facuMH (approved):
This PR replaces a few instances of rules relate to jump for calls to `rulesFor`, Could you add to the PR the impact this change has in number of tests generated ?

HerbertJordan (approved):
Looks good, thanks for identifying the issues and fixing the rules!

PR #145, opened 2025-09-12
Configure MSRV and fix new clippy lints

LorenzSchueler:
This PR sets the minimum supported Rust version (MSRV) in `Cargo.toml` to `1.86`.
This does not change the MSRV, it just configures it so that if an older version of Rust is used, once gets a more helpful error message.

This PR also fixes clippy lints that have been introduced in newer Rust versions (up to `1.89`).

Those lints do not break CI because Rust is already pinned to `1.86` in CI.

### DIFF
--- a/rust/Cargo.toml
+++ b/rust/Cargo.toml
@@ -12,6 +12,7 @@ members = [
 name = "evmrs"
 version = "0.1.0"
 edition = "2024"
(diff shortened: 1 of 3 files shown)


simonlechner (approved):
The comment regarding function pointer equality sounds reasonable.

